PR #45, opened 2019-11-10
Generate roblox standard library automatically if it does not exist

### LINKED ISSUE #42: Selene should auto-generate standard libraries if possible
If a user tries to use a standard library, e.g. `roblox`, and there's no such associated `roblox.toml`, selene should check if it has a procedure to auto-generate that file. If it does, it should do so and write the contents to `roblox.toml` for future runs. Perhaps include a message at the end telling the user how to manually invoke that procedure again if they need to.

Kampfkarren:
Closes #42


Kampfkarren:
Also includes...an identical unscoped_variables for some reason? Whatever...

PR #64, opened 2019-11-26
Add string.split to roblox std

evaera:
Second param is actually optional but idk how to denote that


Kampfkarren:
Yeah this isn't right. This means that the only acceptable usage will be `string.split("string")` (and no other value). Take a look at the docs for the standard library or the built in ones for an example of how it should look.

evaera:
Just fix it yourself loser

Kampfkarren:
Looks good! Can you quickly update the CHANGELOG and I can merge this in? :)

Kampfkarren:
Perfect!

PR #85, opened 2020-01-25
Add tuple argument to xpcall

### LINKED ISSUE #78: xpcall doesn't accept tuple args (parameters for the first argument, the function being called in protected mode))
image

image

dekkonot:
Should resolve #78


Kampfkarren:
Awesome, thanks!

PR #101, opened 2020-04-22
Add missing table functions

NavigationBar:
This PR adds 3 missing table functions to the Roblox standard library.


Kampfkarren:
All of these are deprecated, so I don't think we want to add these to the standard library.

PR #106, opened 2020-04-25
Clear lints when a file is deleted

### LINKED ISSUE #92: VSCode extension keeps errors and warnings from deleted files
If a file with errors in it is deleted, those errors will remain in the Problems menu until the window is reloaded.

LouieK22:
Closes #92

### DIFF
--- a/selene-vscode/src/extension.ts
+++ b/selene-vscode/src/extension.ts
@@ -121,9 +121,16 @@ export async function activate(context: vscode.ExtensionContext) {
         diagnosticsCollection.set(document.uri, diagnostics)
     }
 
+    async function clearLints(documentUri: vscode.Uri) {
+        diagnosticsCollection.set(documentUri, [])
+    }
+
     vscode.workspace.onDidSaveTextDocument(lint)
     vscode.workspace.onDidOpenTextDocument(lint)
     vscode.workspace.onDidChangeTextDocument(event => lint(event.document))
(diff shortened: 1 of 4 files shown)


selene-vscode/src/extension.ts
```diff
@@ -121,9 +121,16 @@ export async function activate(context: vscode.ExtensionContext) {
...
+    async function clearLints(documentUri: vscode.Uri) {
+        diagnosticsCollection.set(documentUri, [])
+    }
+
     vscode.workspace.onDidSaveTextDocument(lint)
     vscode.workspace.onDidOpenTextDocument(lint)
     vscode.workspace.onDidChangeTextDocument(event => lint(event.document))
+    vscode.workspace.onWillDeleteFiles(event => {
+        event.files.forEach(clearLints);
+    })
```
Kampfkarren, line 133: I think having the `clearLints` function separate from the only place it's being called hurts readability. This should just be changed to a `for (const ... of ...)` loop.

LouieK22:
Dependency updates are needed since the API for onWillDeleteFiles didn't exist on the current version.

Kampfkarren:
Makes sense then, will test this on my own machine tomorrow.

Kampfkarren:
image
Doesn't seem to work for me?

Kampfkarren:
Can you add a CHANGELOG entry under an `Unreleased` header?

Kampfkarren:
Thanks!

PR #109, opened 2020-04-30
Let diagnostics hold their own string

tversteeg:
By making the diagnostic hold their own strings instead of having a static reference, memory doesn't have to be leaked.

### DIFF
--- a/selene-lib/src/lib.rs
+++ b/selene-lib/src/lib.rs
@@ -167,7 +167,7 @@ macro_rules! use_rules {
                 })
             }
 
-            pub fn test_on(&self, ast: &Ast<'static>) -> Vec<CheckerDiagnostic> {
+            pub fn test_on(&self, ast: &Ast) -> Vec<CheckerDiagnostic> {
                 let mut diagnostics = Vec::new();
 
                 macro_rules! check_rule {
--- a/selene-lib/src/rules.rs
+++ b/selene-lib/src/rules.rs
@@ -40,7 +40,7 @@ pub trait Rule {
     fn new(config: Self::Config) -> Result<Self, Self::Error>
     where
         Self: Sized;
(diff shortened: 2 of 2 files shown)


Kampfkarren:
I initially used this because I thought owning strings would lead to higher memory usage when using identical strings, my testing shows LLVM should intern them. Is it possible to see which takes up more memory?

tversteeg:
I guess but the problem is that the only way to convert a string to a `'static` string is to leak memory..

Kampfkarren:
The trick is that there's no conversion taking place--the static strings are embedded into the code already.

Example:

https://github.com/Kampfkarren/selene/blob/master/selene-lib/src/rules/almost_swapped.rs#L31-L32

tversteeg:
When I try to do this:

```rust
let checker = Checker::new(CheckerConfig::default(), standard_library).unwrap();

let ast = full_moon::parse(contents).unwrap();

let diagnostics = checker.test_on(&ast);
```

It complains about the lifetime of `Ast` not being static due to the strings..

Kampfkarren:
`Ast<'static>` doesn't mean the strings inside are of static lifetime, it just means they're all owned.

Kampfkarren:
Does using `&ast.owned()` instead fix it?

tversteeg:
That gives me this error:

```
let diagnostics = checker.test_on(&ast.to_owned());
                                  ^^^^^^^^^^^^^^^ lifetime `'static` required
```

Kampfkarren:
Not `to_owned()`, `.owned()` (from full_moon::owned::Owned) -- it's a custom trait full-moon has that forces ownership of all the `Cow`s, making them `Cow<'static>`, though it doesn't leak memory.

tversteeg:
Oh sorry, that works, thanks! Maybe it should be added to the documentation because I wouldn't be able to figure this out by myself without your help :)

Kampfkarren:
The selene-lib documentation is fairly sparse, I agree, definitely want to give it more TLC.

PR #111, opened 2020-05-07
Add ypcall

noahwillcrow:


Kampfkarren:
`ypcall` is deprecated, `pcall` does the exact same thing now. On principle, we don't accept deprecated functionality. Is there any reason you are using this?

noahwillcrow:
It's in `GameAnalytics` SDK and if we can't ignore files I want to still be able to run CI. I'm happy to keep ypcall defined just in my repo's version of things, though, if it is necessary.

Kampfkarren:
`pcall` does the exact same thing as `ypcall` now, so you should just switch over to that.

noahwillcrow:
Hmm. Okay. I'll look into sending a PR over to GameAnalytics and for now change my copy locally.

PR #120, opened 2020-06-21
Change how arguments passed are counted

### LINKED ISSUE #103: assert with a function call
Selene gives a lint when using assert without 2 parameters, but it does not consider that a function call can return multiple values.

It's usually ok but when using [`t`](https://github.com/Roblox/t) it does not like this:
```lua
assert(t.number(foo))
```

The message:
```
standard library function `assert` requires 2 parameters, 1 passed
```

Also, technically `assert` can accept one argument, but I like the idea that selene could suggest (with maybe just a warning) that `assert` is missing the second argument.

jeparlefrancais:
Closes #103 

I added a .gitattributes so that on checkout, git does not change line endings for the `.stderr` files. Otherwise it makes all the lint tests fail if someone has `core.autocrlf` on. I'm not that familiar with that kind of stuff but from the test I made I think it works.

This is an attempt at fixing this use case:
```lua
assert(t.string(variable))
```
The standard library enforces that `assert` is called with two arguments, which I think is a good idea. But it does not work well when using a library like [`t`](https://github.com/Roblox/t), that uses function that returns the two values.

#### What I did
The way I solved it is by checking if the last value passed to the function is a `...` or a function call. Then if it is, it updates the minimum amount of values passed to the function. So it is still possible to know if there are too much arguments.

#### Another option
Another way I thought about doing it was to change the `get_argument_type` function to return `ArgumentType::Vararg` in the function call case. After that it's possible to just check if the last argument type is `Vararg` to determine if the max number of arguments is unknown or not.

I don't know if that could have a larger impact on other stuff so that's why I went with what I did.

#### Bad side of the fix
The sad thing is that it allows weird things to pass. Like this for example:
```
print( type(print(), print()) )
```
Which is technically correct, but horrible.

Let me know your thoughts!

### DIFF
--- a/selene-lib/tests/lints/standard_library/assert.lua
+++ b/selene-lib/tests/lints/standard_library/assert.lua
@@ -1,0 +1,4 @@
+assert(true, 'message')
+assert(call())
+assert(true)
+assert(call(), 'this is ok')
--- a/selene-lib/tests/lints/standard_library/assert.lua
+++ b/selene-lib/tests/lints/standard_library/assert.lua
@@ -1,4 +1,8 @@
 assert(true, "message")
 assert(call())
-assert(true)
 assert(call(), "this is ok")
+assert(...)
+
+assert(true)
+assert(true, "message", call())
+assert(true, "message", ...)
--- a/selene-lib/tests/lints/standard_library/assert.lua
+++ b/selene-lib/tests/lints/standard_library/assert.lua
@@ -1,4 +1,4 @@
-assert(true, 'message')
+assert(true, "message")
 assert(call())
 assert(true)
-assert(call(), 'this is ok')
+assert(call(), "this is ok")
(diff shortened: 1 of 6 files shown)


Kampfkarren:
> The sad thing is that it allows weird things to pass. Like this for example:
> 
> print( type(print(), print()) )

What causes this? This seems like something that should be simple to block.

selene-lib/tests/lints/standard_library/assert.lua
```diff
@@ -0,0 +1,4 @@
+assert(true, 'message')
```
Kampfkarren, line 1: Can you replace this and the other one with double quotes just to be consistent?

jeparlefrancais:
> What causes this? This seems like something that should be simple to block.

Yes, you are right! I changed it so it assumes that a function call or `...` give at least one argument. I just added a flag to know if there are potentially more arguments.

Kampfkarren:
Can you add that to the tests now that it works?

Kampfkarren:
Thanks a bunch!

PR #132, opened 2020-07-19
Add RaycastParams

### LINKED ISSUE #119: Selene says RaycastParams is not defined
I was going to use the new raycasting API but selene doesn't seem to recognize it.

image

MicahHinckley:
This works, however, I am unsure if this is the correct way to define a constructor that doesn't have parameters.

Closes #119.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,13 @@
 # Changelog
 This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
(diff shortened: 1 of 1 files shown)


Kampfkarren:
You based this commit off your other PR, so the logs and such are messy, please fix it to rebase off master, make your changes again, and force push.

Also, this is not the way to make no parameter functions, this is:

```toml
[RaycastParams.new]
args = []
```

PR #141, opened 2020-07-23
Fix a typo with comments_count

### LINKED ISSUE #139: empty_if comments_count not working
My guess would be that the option does not work for `elseif`s? Edit: nvm found another one.

Function

```lua
thumbnail:SetScript("OnUpdate", function(self, elapsed)
    self.elapsed = self.elapsed + elapsed;
    if(self.elapsed < 0.5) then
      t1:SetVertexColor(1,0,0,1);
    elseif(self.elapsed < 1.5) then
      t1:SetVertexColor(1,1,1,1);
    elseif(self.elapsed < 3) then
      -- do nothing
    else
      self.elapsed = self.elapsed - 3;
    end
end);
```

Config:

```toml
[config]
empty_if = { comments_count = false }
unused_variable = { allow_unused_self = true }
```

Warning:

```
warning[empty_if]: empty elseif block

     ┌── WeakAurasTemplates/TriggerTemplates.lua:443:9 ───
     │
 443 │ ╭         elseif(self.elapsed < 3) then
 444 │ │           -- do nothing
 445 │ │         else
     │ ╰────────^
     │
```

Bonus:

```
     ┌── WeakAurasOptions/WeakAurasOptions.lua:589:5 ───
     │
 589 │ ╭     if(data.parent) then
 590 │ │     -- Do nothing; children will be added later
 591 │ │     elseif not(loaded[id]) then
 592 │ │       tinsert(to_sort, id);
 593 │ │     end
     │ ╰───────^
     │
```

Stanzilla:
Fixes #139

### DIFF
--- a/docs/src/cli/configuration.md
+++ b/docs/src/cli/configuration.md
@@ -35,7 +35,7 @@ Where the value is whatever the special configuration of that rule is. You can l
 
 ```toml
 [config]
-empty_if = { comments_count = false }
+empty_if = { comments_count = true }
 ```
 
 ## Setting the standard library


Kampfkarren:
Thanks!

PR #163, opened 2020-10-04
Function calls can now be used as the last parameter of a function call without checking for number of arguments passed

### LINKED ISSUE #75: Parameter count should not be inferred if a function is used as an argument
The following errors, though it should not:

`math.max(unpack(lootRarities))`

Kampfkarren:
Fixes #75.


Kampfkarren:
image

☹️

Issue #114, opened 2020-06-02
standard library function `Instance.new` requires 1 parameters, 2 passed


ReinitiaIized:
**What's happening:**
selene is throwing a complaint regarding the second parameter of Instance.new not being valid
![unknown 1](https://user-images.githubusercontent.com/20674500/83531702-f2088500-a4b2-11ea-9e31-87ef7443f727.png)

**What should happen**
selene should not be complaining as it is a valid parameter

Kampfkarren:
This is intentional. You should not be using the second parameter of Instance.new as it is not idiomatic at best, ill performant at worst.

ReinitiaIized:
I actually remembered the use case for this. 

Say I have a need to use a BindableEvent for suspending the calling thread, however I don't need to change any of its properties beyond parenting it, then destroying it after its been used. This is a case where I argue a **strong** warning should be pushed, stating performance reasons, but not outright calling it an error.

Kampfkarren:
You should still set the parent property instead for consistency.

PR #164, opened 2020-10-14
Fix Instance.new arguments

Minstrix:
Code_FkuLUC6VbK

### DIFF
--- a/selene/src/roblox/base.toml
+++ b/selene/src/roblox/base.toml
@@ -338,6 +338,9 @@ type = "..."
 [[Instance.new.args]]
 type = "string" # This is changed at generate time
 
+[[Instance.new.args]]
+type = "any"
+
 [[math.clamp.args]]
 type = "number"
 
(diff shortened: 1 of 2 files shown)


Jaguar-515:
https://github.com/Kampfkarren/selene/issues/114#issuecomment-637844358

Kampf stated in an old issue that he won't fix this as you **shouldn't** pass a parent argument in Instance.new().

On another note, could we make a lint for this? Seeing that sometimes this arises up occasionally, we should start to make a lint for this that allow/disallow using two arguments for Instance.new().

Minstrix:
Ah, that's my bad then, I thought I'd searched the issues but must've only checked for past pull requests. Personally, I still don't think this should be marked as incorrect usage for no more than consistency. But, if Kampf is firm on this, making an allow/disallow lint might be a good compromise, although I'd have no idea how to go about doing that.

Kampfkarren:
I'd rather make this its own lint, or buff the standard library lint to allow for cases like this.

selene/src/roblox/base.toml
```diff
@@ -338,6 +338,9 @@ type = "..."
 [[Instance.new.args]]
 type = "string" # This is changed at generate time
 
+[[Instance.new.args]]
+type = "any"
```
Kampfkarren, line 342: This would break every case of `Instance.new("Class")` by the way, since you're not doing `required = false`.

Issue #99, opened 2020-04-14
Silence unbalanced_assignments when right hand side is ...


Kampfkarren:
```lua
local a, b = ...
```

...should not lint.

PR #168, opened 2020-10-21
Remove lint when use unbalanced assignment with ellipsis

rodrigocam:
Remove lint when using unbalanced assignments with ellipsis, e.g. ```local a, b, c = ...```

resolves https://github.com/Kampfkarren/selene/issues/99

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,10 +10,10 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 ### Changed
 - setmetatable no longer requires a second argument.
 - `allow_unused_self` is now toggled on for `unused_variable` by default.
-- unbalanced assignments with ellipsis no longer lints
 
 ### Fixed
 - Using a function call as the last argument in a function will silence lint for not passing enough parameters. This means, for example, `math.max(unpack(numbers))` will no longer error.
+- Using an ellipsis on the right side of unbalanced assignments no longer lints.
 
 ## [0.8.0] - 2020-08-24
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 ### Changed
 - setmetatable no longer requires a second argument.
 - `allow_unused_self` is now toggled on for `unused_variable` by default.
+- unbalanced assignments with ellipsis no longer lints
 
 ### Fixed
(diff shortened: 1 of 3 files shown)


selene-lib/src/rules/unbalanced_assignments.rs
```diff
@@ -113,6 +113,21 @@ fn expression_is_nil(expression: &ast::Expression) -> bool {
...
+                *symbol.token_type() == TokenType::Symbol { symbol: Symbol::Ellipse} 
+            } else {
+                false
+            }
+            
+        },
+
+        _ => false
+    }
+}
```
Kampfkarren, line 129: Can you collapse this match into another `if let` and do `return *symbol.token_type() == TokenType::Symbol { symbol: Symbol::Ellipse }` when they pass? Then, at the end, have `false` as a fail safe.

Kampfkarren:
CHANGELOG as well, please.

selene-lib/src/rules/unbalanced_assignments.rs
rodrigocam: Do you mean?

```    
if let ast::Expression::Value{value, ..} = expression {
    if let ast::Value::Symbol(symbol) = &**value {
        *symbol.token_type() == TokenType::Symbol { symbol: Symbol::Ellipse }
    } else {
        false
    }
} else {
    false
}

```

Kampfkarren: Yep, though minor style changes:

```rs
if let ast::Expression::Value{value, ..} = expression {
    if let ast::Value::Symbol(symbol) = &**value {
        return *symbol.token_type() == TokenType::Symbol { symbol: Symbol::Ellipse };
    } 
} 

false
```

rodrigocam: Done!

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 ### Changed
 - setmetatable no longer requires a second argument.
 - `allow_unused_self` is now toggled on for `unused_variable` by default.
+- unbalanced assignments with ellipsis no longer lints
```
Kampfkarren, line 13: Last nitpick before merge--this was a bug, so it should be under `Fixed`, as well as with a period at the end.

rodrigocam: No problem! Done.

Kampfkarren:
Perfect, thanks!

PR #176, opened 2020-11-09
Update build.yml

### LINKED ISSUE #151: Foreman installs the selene-light executable instead of selene on Linux
To reproduce, use [foreman](https://github.com/Roblox/foreman) to install the latest version of selene on Linux; attempting to use Roblox features such as `generate-roblox-std` will fail. Foreman selects the executable at the first index of a release archive when installing, so this might be happening because `selene-light` was zipped into `selene-0.8.0-linux.zip` before `selene`.

y3th:
Closes #151.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -22,11 +22,16 @@ jobs:
         cd selene
         cargo build --locked --release --verbose --no-default-features
         Move-Item ..\target\release\selene.exe .\release\selene-light.exe
-    - name: Upload
+    - name: Upload selene
       uses: actions/upload-artifact@v1
       with:
         name: selene-windows
-        path: ./selene/release
+        path: ./selene/release/selene.exe
+    - name: Upload selene-light
+      uses: actions/upload-artifact@v1
+      with:
+        name: selene-light-windows
+        path: ./selene/release/selene-light.exe
   build_mac:
     runs-on: macos-latest
     steps:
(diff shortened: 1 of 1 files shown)


Kampfkarren:
I'm confused, what exactly is this doing and how does this solve that issue? I think that issue is already solved even now that with Foreman I'm just dropping the zip files directly.

y3th:
Foreman still ends up installing selene-light.

image

All this PR does is change the order in which the light version of selene and the full version of selene are built. Building the full version of selene first should solve the issue. 

Honestly, it'd be better if Foreman allowed you to choose an executable by name in an artifact.

Kampfkarren:
I don't think it matters what the order is since it's just a zip file, @LPGhatguy can you bring any insight to this?

LPGhatguy:
Selene's releases have more than one executable file in the same zip. Foreman just picks the first executable it finds, which looks like it might be the light one. I don't think this fix is very robust.

The light and full builds of Selene should probably be separate artifacts. Bundling both in the same zip is very strange to me, as you likely only want one or the other, and this blows up the size of the archive.

Kampfkarren:
> Selene's releases have more than one executable file in the same zip. 

Correct.

I can see separating them into different zips. CC @y3th on if you want to do that.

.github/workflows/build.yml
```diff
@@ -11,63 +11,78 @@ jobs:
...
-      run: |
-        cd selene
-        cargo build --locked --release --verbose --no-default-features
-        New-Item release -ItemType "directory"
-        Move-Item ..\target\release\selene.exe .\release\selene-light.exe
     - name: Build (All features)
       run: |
         cd selene
         cargo build --locked --release --all-features
+        New-Item release -ItemType "directory"
```
Kampfkarren, line 18: If we're going to be just shipping out the executables directly, I don't think we need to be making directories anymore--you can just send straight from target/release.

Kampfkarren (changes requested):

.github/workflows/build.yml
y3th: Done! I would've preferred to keep the build steps and the upload steps grouped, but this seems faster.

Kampfkarren (approved):

PR #182, opened 2020-11-19
Roblox Lua allows error first argument to be a table

### LINKED ISSUE #180: Roblox error can accept any type for first argument
Using selene 0.9.2, I'm seeing this output:
```
error[incorrect_standard_library_use]: use of standard_library function `error` is incorrect
    ┌─ StackFrame.lua:113:10
    │
113 │               error({
    │ ╭───────────────────^
114 │ │                 stack = debug.traceback(),
115 │ │             })
    │ ╰─────────────^ expected `string`, received `table`
```

The (correct) error definition for Lua 5.1 is here, https://github.com/Kampfkarren/selene/blob/master/selene-lib/default_std/lua51.toml#L32
but Roblox Lua allows for a table to be given for the first argument as well. 

It looks like the place to insert this might be here: https://github.com/Kampfkarren/selene/blob/master/selene/src/roblox.rs#L49
but I'm not sure. If someone can point me to where the right place for the enhancement is, I'll submit a PR.

matthargett:
* Add an entry to the Roblox config that allows more than strings for the first argument to error()

Closes #180

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 ### Added
 - Added inline lint filtering, read [the documentation](https://kampfkarren.github.io/selene/usage/filtering.html) for more information.
 - More errors now set the exit code.
+- Added support for error({any}) to the Roblox standard library.
 
 ## [0.9.2] - 2020-11-06
 ### Changed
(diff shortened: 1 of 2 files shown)


Kampfkarren:
Thanks! Can you put this in the CHANGELOG as well?

Kampfkarren:
Thanks!

PR #187, opened 2020-12-12
Fix minor spelling and formatting in docs

KyleMit:
Nothing super helpful here, but hopefully little bit of cleanup.  Just typos in the docs and formatting on markdown header indentation levels.

Only other spelling tidbit is `parenthese` which should be `parenthesis` if singular or `parentheses` if plural, but probably not worth changing since it's part of the external lint config rules


Kampfkarren:
Good catch, thanks!

PR #199, opened 2020-12-31
Fix diagnostic positions when using Unicode characters

JohnnyMorganz:
This PR should fix the problem where diagnostics were incorrectly positioned in the VSCode extension when there are Unicode characters in the file

VSCode only accepts character offsets, however selene will return byte offsets for its start and end positions. This PR implements a conversion from byte offsets to character positions

### DIFF
--- a/selene-vscode/src/extension.ts
+++ b/selene-vscode/src/extension.ts
@@ -26,10 +26,24 @@ enum Severity {
     Warning = "Warning",
 }
 
+function byteToCharOffset(document: vscode.TextDocument, byteOffset: number) {
+    const text = document.getText();
+    let currentOffset = 0;
+    // Iterate through each character in the string
+    for (let i = 0; i < byteOffset; i++) {
+        // Calculate the current byte offset we have reached so far
+        currentOffset += Buffer.byteLength(text[i], "utf-8");
+        if (currentOffset >= byteOffset) {
+            return i + 1
+        }
+    }
+    return currentOffset
+}
+
 function labelToRange(document: vscode.TextDocument, label: Label): vscode.Range {
     return new vscode.Range(
-        document.positionAt(label.span.start),
-        document.positionAt(label.span.end),
+        document.positionAt(byteToCharOffset(document, label.span.start)),
+        document.positionAt(byteToCharOffset(document, label.span.end)),
     )
 }
 
--- a/selene-vscode/src/extension.ts
+++ b/selene-vscode/src/extension.ts
@@ -26,24 +26,32 @@ enum Severity {
     Warning = "Warning",
 }
 
-function byteToCharOffset(document: vscode.TextDocument, byteOffset: number) {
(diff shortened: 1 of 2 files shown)


Dekkonot:
I'm worried about the performance impact of this, especially on larger files, where the extension already has some performance issues (I couldn't find a relevant issue despite talking about it with Kampf before).

I know the timeline for PRs is usually more than a day but just in case I'm asking that this not be merged until we can actually check what the impact will be. I'll try to take a look tonight.

selene-vscode/src/extension.ts
```diff
@@ -26,10 +26,24 @@ enum Severity {
     Warning = "Warning",
 }
 
+function byteToCharOffset(document: vscode.TextDocument, byteOffset: number) {
+    const text = document.getText();
```
Kampfkarren, line 30: No semicolons

selene-vscode/src/extension.ts
```diff
@@ -26,10 +26,24 @@ enum Severity {
...
+    let currentOffset = 0;
+    // Iterate through each character in the string
+    for (let i = 0; i < byteOffset; i++) {
```
Kampfkarren, line 33: Use `byte` instead of `i`.

Kampfkarren (changes requested):
I think it's worth looking to see how other extensions solve this issue as well @Dekkonot

selene-vscode/src/extension.ts
JohnnyMorganz: This is an iteration over the file characters, would `char` be more suitable here?

Kampfkarren: Sure, both are more descriptive than `i` (though you should prefer `character`).

Kampfkarren: I'd use something like `charOffset` though, I'm not sure what this is but I know it's a number, and a variable named just `char` should refer to a string.

Kampfkarren:
@Dekkonot Any update?

Dekkonot:
> @Dekkonot Any update?

I've been busy with irl things for the past few days so I haven't had time to take a look. I should know better than to say things like "I'll try to do it tonight"!

I have some free time tonight that I might use but worst-case it'll be in a few days because then I have all day free. If one of you (cc @JohnnyMorganz) wants to benchmark it, I'm okay with that though.

One thought is that it might be faster/easier to do this on the Rust side of things and have a CLI flag for it rather than doing it in the extension. That has the added benefit of fixing it globally rather than just in the vscode extension, because I'm sure other editors run into the same problem.

Dekkonot:
I haven't ran any numbers but it seems to perform about the same as before. I don't have a bunch of unicode in any of my code (I'm lucky enough to primarily write in English) but when I added random unicode characters I didn't notice any differences.

It gets my vote of approval at any rate @Kampfkarren.

Kampfkarren:
:+1: Sound

PR #211, opened 2021-03-19
Remove use of `Ast::owned`.

tomprince:
Since we are using the AST before the source gets deallocated, there is no reason to copy al of the source.


Kampfkarren:
I'm... sort of surprised this compiled with such little changes! I remember this being a code complexity thing.

PR #218, opened 2021-03-25
Standard libraries now allow the creation of callable metatables

### LINKED ISSUE #170: Impossible to define callable metatables
TestEZ's expect is a novel example of this:

```lua
return function()
    it("", function()
        expect.extend({})

        expect(1).to.be.ok()
    end)
end
```

I would expect this to work:
```
[[expect.args]]
required = true
type = "any"

[[expect.extend.args]]
required = true
type = "any"
```

however the two definitions conflict, and you'll only be able to lint the first, but not the other.

Kampfkarren:
Closes #170

This is a bit of a weird fix, and worked first try, so I'm very skeptical of it. Will run tests on existing codebases and such later.


Kampfkarren:
generate-roblox-std outputs from before and after are identical, except now for some reason args is after method. This is fine, since these shouldn't be in git anyway.

Kampfkarren:
!selenology

PR #220, opened 2021-03-25
Use `vscode.workspace.fs` where possible

JohnnyMorganz:
This PR makes uses of `vscode.workspace.fs` instead of the Node.js `fs` and `path` module, where possible.
I had to bump the VSCode engine version to latest, to use some newer `fs`-related methods.

We now pass around `vscode.Uri`s where possible, instead of file system paths (however we *do* still need them).
Using this allows us to handle remote/other file uris.

One thing that wasn't changed, was using `fsWriteFileAtomic` to download the extension. `vscode.workspace.fs.writeFile` is an alternative, but I'm not sure if its suited for writing such a large binary, and it doesn't support piping the stream.

### DIFF
--- a/selene-vscode/src/util.ts
+++ b/selene-vscode/src/util.ts
@@ -84,11 +84,11 @@ function getSeleneFilenamePattern(): RegExp {
 
 async function fileExists(filename: vscode.Uri): Promise<boolean> {
     try {
-        await vscode.workspace.fs.stat(filename);
-        return true;
+        await vscode.workspace.fs.stat(filename)
+        return true
     } catch (err) {
         // If an error was thrown, the file was not found
-        return false;
+        return false
     }
 }
 
@@ -133,7 +133,7 @@ export async function getSelenePath(storagePath: vscode.Uri): Promise<vscode.Uri
         return vscode.Uri.file(settingPath)
     }
 
-    const downloadPath = vscode.Uri.joinPath(storagePath, getSeleneFilename());
+    const downloadPath = vscode.Uri.joinPath(storagePath, getSeleneFilename())
     if (await fileExists(downloadPath)) {
         return downloadPath
     }
--- a/selene-vscode/src/util.ts
+++ b/selene-vscode/src/util.ts
@@ -87,8 +87,7 @@ async function fileExists(filename: vscode.Uri): Promise<boolean> {
         await vscode.workspace.fs.stat(filename)
         return true
     } catch (err) {
-        // If an error was thrown, the file was not found
-        return false
+        return err != vscode.FileSystemError.FileNotFound
     }
 }
 
--- a/selene-vscode/src/util.ts
+++ b/selene-vscode/src/util.ts
@@ -143,7 +143,7 @@ export async function ensureSeleneExists(storagePath: vscode.Uri) {
     const path = await getSelenePath(storagePath)
 
     if (path === undefined) {
-        await vscode.workspace.fs.createDirectory(storagePath);
+        await vscode.workspace.fs.createDirectory(storagePath)
         return downloadSelene(storagePath)
     } else {
         if (!await fileExists(path)) {
--- a/selene-vscode/src/util.ts
+++ b/selene-vscode/src/util.ts
@@ -1,6 +1,4 @@
-import * as fs from "fs"
 import * as os from "os"
-import * as path from "path"
 import * as selene from "./selene"
 import * as requestNative from "request"
 import * as request from "request-promise-native"
@@ -84,13 +82,19 @@ function getSeleneFilenamePattern(): RegExp {
     }
 }
 
-async function fileExists(filename: string): Promise<boolean> {
-    return fs.promises.stat(filename)
-        .then(() => true)
-        .catch((error) => error.code !== "ENOENT")
+async function fileExists(filename: vscode.Uri): Promise<boolean> {
+    try {
+        await vscode.workspace.fs.stat(filename);
+    } catch (err) {
+        if (err == vscode.FileSystemError.FileNotFound) {
+            return false;
+        }
+    }
+
+    return true;
 }
 
-export async function downloadSelene(directory: string) {
+export async function downloadSelene(directory: vscode.Uri) {
     vscode.window.showInformationMessage("Downloading Selene...")
 
     const filename = getSeleneFilename()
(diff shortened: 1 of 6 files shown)


selene-vscode/src/util.ts
```diff
@@ -84,13 +82,17 @@ function getSeleneFilenamePattern(): RegExp {
...
-        .then(() => true)
-        .catch((error) => error.code !== "ENOENT")
+async function fileExists(filename: vscode.Uri): Promise<boolean> {
+    try {
+        await vscode.workspace.fs.stat(filename);
+        return true;
+    } catch (err) {
+        // If an error was thrown, the file was not found
+        return false;
+    }
```
Kampfkarren, line 92: Why change this from what there was before? `Promise.catch` should work perfectly fine.

Kampfkarren: Plus, what was there before captures in the case of *other* failures that can happen during I/O. I'd be surprised if vscode.workspace.fs.stat didn't provide something similar, if not identical to `error.code`.

selene-vscode/src/util.ts
```diff
@@ -125,23 +127,23 @@ export async function downloadSelene(directory: string) {
...
-export async function getSelenePath(storagePath: string): Promise<string | undefined> {
+export async function getSelenePath(storagePath: vscode.Uri): Promise<vscode.Uri | undefined> {
     const settingPath = vscode.workspace.getConfiguration("selene").get<string | null>("selenePath")
     if (settingPath) {
-        return settingPath
+        return vscode.Uri.file(settingPath)
     }
 
-    const downloadPath = path.join(storagePath, getSeleneFilename())
+    const downloadPath = vscode.Uri.joinPath(storagePath, getSeleneFilename());
```
Kampfkarren, line 136: Are these not prototypes, does `storagePath.joinPath` not work?

selene-vscode/src/util.ts
```diff
@@ -125,23 +127,23 @@ export async function downloadSelene(directory: string) {
...
     }
 }
 
-export async function ensureSeleneExists(storagePath: string) {
+export async function ensureSeleneExists(storagePath: vscode.Uri) {
     const path = await getSelenePath(storagePath)
 
     if (path === undefined) {
-        await fs.promises.mkdir(storagePath, { recursive: true })
+        await vscode.workspace.fs.createDirectory(storagePath);
```
Kampfkarren, line 146:
```suggestion
        await vscode.workspace.fs.createDirectory(storagePath)
```

selene-vscode/src/util.ts
Kampfkarren: Also, don't use semicolons to be consistent with the rest of the code. I need to remember to put Prettier on here at some point.

JohnnyMorganz: Doesn't look like it - https://code.visualstudio.com/api/references/vscode-api#Uri.

Apologies for the semicolons, I tried to remove them where possible, but I guess it's just muscle memory 😛

selene-vscode/src/util.ts
JohnnyMorganz: `vscode.workspace.fs.stat(filename)` returns a `Thenable`, so it only has a `.then()`, but no `.catch()` - it will just throw the error.
For your second point, I could change it to
```suggestion
    try {
        await vscode.workspace.fs.stat(filename);
        return true;
    } catch (err) {
        return err != vscode.FileSystemError.FileNotFound;
    }
```

Kampfkarren: Ugh, that's annoying. Yes, that's better than a blanket catch.

selene-vscode/src/util.ts
```diff
@@ -84,13 +82,16 @@ function getSeleneFilenamePattern(): RegExp {
...
-async function fileExists(filename: string): Promise<boolean> {
-    return fs.promises.stat(filename)
-        .then(() => true)
-        .catch((error) => error.code !== "ENOENT")
+async function fileExists(filename: vscode.Uri): Promise<boolean> {
+    try {
+        await vscode.workspace.fs.stat(filename)
+        return true
+    } catch (err) {
+        return err != vscode.FileSystemError.FileNotFound
```
Kampfkarren, line 90: !==

selene-vscode/src/util.ts
```diff
@@ -84,13 +82,16 @@ function getSeleneFilenamePattern(): RegExp {
...
+    }
 }
 
-export async function downloadSelene(directory: string) {
+export async function downloadSelene(directory: vscode.Uri) {
```
Kampfkarren, line 94: Are you sure this should be using the vscode fs? It should only ever be downloading to an AppData folder or similar, seems like the standard fs stuff would be fine.

JohnnyMorganz: The value originally passed to this was `ExtensionContext.globalStoragePath`, which was a string containing the fs path. This was deprecated in favour of `ExtensionContext.globalStorageUri`. I felt it would probably be wise to utilise `globalStorageUri` itself as much as possible (also allows us to use `vscode.Uri.joinPath`), and then finally use the normal filesystem path at the end when writing using `fsWriteFileAtomic` (see: line 103).

You may also want to consider removing `fsWriteFileAtomic` completely some time in the future with the vscode fs write file function - I didn't include changing that in this PR because I was unsure on how suitable it was for the task of writing a stream for a large binary file. Why would you want to do this? To quote https://code.visualstudio.com/api/references/vscode-api#workspace, the vscode fs APIs run outside of the editor process, compared to using the Node.js APIs (which run inside the extension context), so it may perform better [although downloading is still pretty quick for now!]

Kampfkarren: The reason for using vscode stuff makes sense in the context of the global path being that type. The reaon for the atomic file writes is so that if I start to update selene (which can happen when you first open a Lua file), close VS Code, then re-open, it should use the older selene. Before atomic file writes they would be corrupted. All an "atomic file write" is in this system is downloading to a temp location then moving it over I think.

JohnnyMorganz: Mhm, I understand the reasoning behind it - I'm not sure the VSCode APIs are suitable for what you want right now, so it should be left as-is 👍

JohnnyMorganz:
Just realised Prettier seems to have acted up again with the diff for `util.ts` - I'll sort it out again when I'm free

Kampfkarren:
Undraft when ready.

JohnnyMorganz:
Apologies, I forgot I had this open!

Kampfkarren:
I forgot we don't actually run Prettier on this codebase yet.

Kampfkarren:
We can definitely write our own atomic write at some point, but this is good enough for now. Can you update the CHANGELOG?

Kampfkarren:
Thanks!

PR #230, opened 2021-04-16
Fix for `Token` not implementing `Node` in full-moon 0.10.

tomprince:


Kampfkarren:
Thanks! We should still probably figure out if Node is the best API though, will probably need to talk to StyLua's maintainers about their usage.

PR #232, opened 2021-04-20
Show more diagnostic information in extension when available

JohnnyMorganz:
Selene provides extra diagnostic information in some cases (mainly parse errors). This information, however, is not currently shown in the extension.
For example:
```lua
function() end
```
What is shown: `unexpected token '('`
Extra information available, but not shown: `expected function name`.
This updates the message to show `unexpected token '(', expected function name`.

This can help in understanding the diagnostic messages better, as it provides useful contextual information.


Kampfkarren:
I knew something was up!

Issue #236, opened 2021-05-03
Choose when to lint via extension


ok-nick:
Although selene is fast, my computer is slow 😞. These features are useful regardless of performance.

A feature to lint only when:
- the file is saved
- a new line is created
- x amount of time after the user has stopped typing

PR #237, opened 2021-05-03
Added `onSave` and `onType` extension configuration

ok-nick:
This feature will allow the user to specify when they want selene to lint via the VS Code extension.

This pr partially checks off [this issue](https://github.com/Kampfkarren/selene/issues/236).

### DIFF
--- a/selene-vscode/package.json
+++ b/selene-vscode/package.json
@@ -40,13 +40,13 @@
                         "null"
                     ],
                     "default": null,
-                    "description": "Specifies the path of selene. If not specified, selene will be automatically downloaded from the GitHub releases."
+                    "description": "Specifies the path of selene. If not specified, selene will automatically download from the GitHub releases."
                 },
                 "selene.run": {
                     "type": "string",
                     "default": "onType",
                     "enum": ["onSave", "onType"],
-                    "description": "Run the linter `onSave` or `onType`."
+                    "description": "Run the linter as you type (onType) or when the file is saved (onSave)."
                 }
             }
         }
--- a/selene-vscode/package.json
+++ b/selene-vscode/package.json
@@ -40,7 +40,7 @@
                         "null"
                     ],
                     "default": null,
-                    "description": "Specifies the path of selene. If not specified, selene will be automatically download from the GitHub releases."
+                    "description": "Specifies the path of selene. If not specified, selene will be automatically downloaded from the GitHub releases."
                 },
                 "selene.run": {
                     "type": "string",
--- a/selene-vscode/package.json
+++ b/selene-vscode/package.json
@@ -40,7 +40,13 @@
                         "null"
                     ],
                     "default": null,
-                    "description": "Specifies the path of selene. If not specified, will automatically download one from the GitHub releases."
+                    "description": "Specifies the path of selene. If not specified, selene will be automatically download from the GitHub releases."
+                },
+                "selene.run": {
+                    "type": "string",
+                    "default": "onType",
+                    "enum": ["onSave", "onType"],
+                    "description": "Run the linter `onSave` or `onType`."
                 }
             }
         }
--- a/selene-vscode/src/extension.ts
+++ b/selene-vscode/src/extension.ts
@@ -166,14 +166,13 @@ export async function activate(context: vscode.ExtensionContext) {
     }
 
     function listenToChange() {
-		switch (vscode.workspace.getConfiguration("selene").get<RunType>("run")) {
-            case RunType.OnSave: {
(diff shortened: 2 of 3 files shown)


selene-vscode/package.json
```diff
@@ -40,7 +40,13 @@
...
                     ],
                     "default": null,
-                    "description": "Specifies the path of selene. If not specified, will automatically download one from the GitHub releases."
+                    "description": "Specifies the path of selene. If not specified, selene will be automatically downloaded from the GitHub releases."
+                },
+                "selene.run": {
+                    "type": "string",
+                    "default": "onType",
+                    "enum": ["onSave", "onType"],
+                    "description": "Run the linter `onSave` or `onType`."
```
Kampfkarren, line 49: Put a better user facing description for this (e.g. "as you're typing")

selene-vscode/src/extension.ts
```diff
@@ -160,9 +160,28 @@ export async function activate(context: vscode.ExtensionContext) {
...
+    let disposable: vscode.Disposable
+    function listenToChange() {
+        if (disposable) {
+            disposable.dispose()
+        }
+
+        const lintOn = vscode.workspace.getConfiguration("selene").get<string>("run")
+        if (lintOn === "onSave") {
+            disposable = vscode.workspace.onDidSaveTextDocument(lint)
+        } else { // onType
```
Kampfkarren, line 172: Use a switch/case and check explicitly, this comment would be unnecessary if the code is checking for it anyway.

selene-vscode/src/extension.ts
```diff
@@ -160,9 +160,28 @@ export async function activate(context: vscode.ExtensionContext) {
         diagnosticsCollection.set(document.uri, diagnostics)
     }
 
-    vscode.workspace.onDidSaveTextDocument(lint)
+    let disposable: vscode.Disposable
+    function listenToChange() {
+        if (disposable) {
+            disposable.dispose()
```
Kampfkarren, line 166: disposable?.dispose()

selene-vscode/src/extension.ts
```diff
@@ -160,9 +160,28 @@ export async function activate(context: vscode.ExtensionContext) {
         diagnosticsCollection.set(document.uri, diagnostics)
     }
 
-    vscode.workspace.onDidSaveTextDocument(lint)
+    let disposable: vscode.Disposable
```
Kampfkarren, line 163: Add `| undefined` to this type, since we can use it before it's initialized.

selene-vscode/src/extension.ts
```diff
@@ -160,9 +160,28 @@ export async function activate(context: vscode.ExtensionContext) {
...
+    function listenToChange() {
+        if (disposable) {
+            disposable.dispose()
+        }
+
+        const lintOn = vscode.workspace.getConfiguration("selene").get<string>("run")
```
Kampfkarren, line 169: Not just string--make a real enum type.

selene-vscode/src/extension.ts
Kampfkarren: On second thought, just dispose it when the configuration is changed. There's no real reason to handle it here.

selene-vscode/src/extension.ts
ok-nick: I'm not sure if this is desirable since I'd have to map each string out to the enum anyways. I could maybe create a function to map it out in `util.ts`?

ok-nick: I'm not sure if this is desirable since I'd have to map each string out to the enum anyways. I could maybe create a function to map it out in util.ts?

Kampfkarren: No, you wouldn't. 

```ts
enum BlaBlaBla {
    OnSave = "onSave",
    OnType = "onType",
}
```

selene-vscode/src/extension.ts
```diff
@@ -161,11 +166,11 @@ export async function activate(context: vscode.ExtensionContext) {
     }
 
     function listenToChange() {
-		switch (vscode.workspace.getConfiguration("selene").get<string>("run")) {
-            case "onSave": {
+		switch (vscode.workspace.getConfiguration("selene").get<RunType>("run")) {
+            case RunType.OnSave: {
```
Kampfkarren, line 170: You don't need these braces, make sure you run prettier.

ok-nick: There is no configuration file for prettier, it ends up formatting each line differently.

Kampfkarren: shame

PR #246, opened 2021-05-17
Disallow `Name` and `Parent` for Roact Host Components

### LINKED ISSUE #245: Lint `Name` property in Roact
Name is never applied from Roact properties.

Should only be used for native elements, not custom ones.

JohnnyMorganz:
Closes #245

### DIFF
--- a/selene-lib/src/rules/roblox_incorrect_roact_usage.rs
+++ b/selene-lib/src/rules/roblox_incorrect_roact_usage.rs
@@ -40,6 +40,17 @@ impl Rule for IncorrectRoactUsageLint {
             ));
         }
 
+        for unused_property in visitor.unused_properties {
+            diagnostics.push(Diagnostic::new(
+                "roblox_incorrect_roact_usage",
+                format!(
+                    "`{}` will never be applied for `{}`",
+                    unused_property.property_name, unused_property.class_name
+                ),
+                Label::new(unused_property.range),
+            ));
+        }
+
         for unknown_class in visitor.unknown_class {
             diagnostics.push(Diagnostic::new(
                 "roblox_incorrect_roact_usage",
@@ -79,6 +90,7 @@ fn is_roact_create_element(prefix: &ast::Prefix, suffixes: &[&ast::Suffix]) -> b
 struct IncorrectRoactUsageVisitor {
     definitions_of_create_element: HashSet<String>,
     invalid_properties: Vec<InvalidProperty>,
+    unused_properties: Vec<InvalidProperty>,
     unknown_class: Vec<UnknownClass>,
 }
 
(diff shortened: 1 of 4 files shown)


selene-lib/tests/lints/roblox_incorrect_roact_usage/roblox_incorrect_roact_usage.stderr
```diff
@@ -16,3 +16,15 @@ error[roblox_incorrect_roact_usage]: `ThisDoesntExist` is not a valid class
 13 │ e("ThisDoesntExist", {})
    │   ^^^^^^^^^^^^^^^^^
 
+error[roblox_incorrect_roact_usage]: `Name` will never be applied for `Instance`
```
Kampfkarren, line 19: You're formatting for a special class name, but it's only showing up as "Instance" here?

Personally, I think we don't need to specify the class name itself, I think just "native Roblox classes" is good enough, and makes it unambiguous as to what the problem is.

selene-lib/src/rules/roblox_incorrect_roact_usage.rs
```diff
@@ -43,10 +43,7 @@ impl Rule for IncorrectRoactUsageLint {
         for unused_property in visitor.unused_properties {
             diagnostics.push(Diagnostic::new(
                 "roblox_incorrect_roact_usage",
-                format!(
-                    "`{}` will never be applied for `{}`",
-                    unused_property.property_name, unused_property.class_name
-                ),
+                format!("`{}` will never be applied", unused_property.property_name),
```
Kampfkarren, line 46: Like I said, let's clarify that this is for native Roblox UI only.

At the same time, we should also show them the correct thing to do when the property is `Name`, as it seems to confuse people on Discord.

JohnnyMorganz: Not entirely sure how I should phrase the help message, without it being too long. Also - since we don't have access to the context surrounding it, its not possible to differentiate between the top-most component (passed to `Roact.mount(component, target, name)`), or if its a child component, where the name is inferred from the key - so it would most likely have to mention both of them, right?

Kampfkarren: It could only mention it when it's inside a table of its own, as a secondary message saying "try: ...", like rustc.

JohnnyMorganz: The visitor currently only visits `FunctionCall` for these - it doesn't know that its inside a table. If I wanted to detect if it was inside a table, wouldn't I have to add a visitor to visit a `TableConstructor`, then go through its fields and recursively check all of them for a `createElement` or equiv. FunctionCall, whilst also keeping the `FunctionCall` visitor to lint other components outside of tables?

Kampfkarren: My approach to this is usually capturing the spans of every table, but if that's too much of a headache then it might not be worth it and can be degerred.

JohnnyMorganz: That does sound like it should work, and shouldn't be too bad to implement, much easier than what I was thinking :) - I'll take a look.

Kampfkarren:
Difference is because it's not merged to upstream yet.

PR #247, opened 2021-05-17
Add mismatched argument count lint

### LINKED ISSUE #39: Lint against too many parameters for defined functions
```lua
local function foo(a, b)
end

foo(1, 2, 3) -- error
```

JohnnyMorganz:
Lints function calls to check that the number of argument counts passed is valid. Also takes into account function definitions which take var args, and function calls where the last argument passed can have a side effect (e.g. `foo(a, b, call())` or `foo(a, b, ...)`)
Closes #39


selene-lib/tests/lints/invalid_type_check/invalid_type_check.lua
```diff
@@ -0,0 +1,5 @@
+if type("string") == "string" then
```
Kampfkarren, line 1: Leftover code?

selene-lib/src/rules/mismatched_arg_count.rs
```diff
@@ -0,0 +1,425 @@
...
+use crate::ast_util::{
+    range,
+    scopes::{ScopeManager, Variable},
+    HasSideEffects,
+};
+use std::{
+    collections::{HashMap, HashSet},
+    convert::Infallible,
+    fmt::{self, Display},
+    usize,
```
Kampfkarren, line 11: Don't need this.

selene-lib/src/rules/mismatched_arg_count.rs
```diff
@@ -0,0 +1,425 @@
...
+    Variable,
+}
+
+impl ParameterCount {
+    /// Calculates the number of required parameters that must be passed to a function
+    pub fn from_function_body(function_body: &ast::FunctionBody) -> Self {
+        let mut necessary_params = 0;
+
+        for parameter in function_body.parameters() {
+            match parameter {
```
Kampfkarren, line 108: If it doesn't warn before we implement the full-moon update that makes this exhaustive, give this a `_` branch.

selene-lib/src/rules/mismatched_arg_count.rs
```diff
@@ -0,0 +1,425 @@
...
+
+struct MismatchedArgCount {
+    parameter_count: ParameterCount,
+    num_provided: PassedArgumentCount,
+    call_range: (usize, usize),
+    function_definition_range: (usize, usize),
+}
+
+#[derive(Clone, Copy, Debug)]
+enum ParameterCount {
```
Kampfkarren, line 93: I sense this might be useful some other day, maybe we should move it into the ast util module?

selene-lib/src/rules/mismatched_arg_count.rs
```diff
@@ -0,0 +1,425 @@
...
+            }
+        }
+
+        Self::Fixed(necessary_params)
+    }
+
+    /// Checks the provided number of arguments to see if it satisfies the number of arguments required
+    /// We will only lint an upper bound. If we have a function(a, b, c) and we call foo(a, b), this will
+    /// pass the lint, since the `nil` could be implicitly provided.
+    pub fn correct_num_args_provided(&self, provided: PassedArgumentCount) -> bool {
```
Kampfkarren, line 126: No need for `&self` if this is `Copy`, then you get to ditch the `*required` and such as well.

selene-lib/src/rules/mismatched_arg_count.rs
```diff
@@ -0,0 +1,425 @@
...
+            ParameterCount::Minimum(_) => true,
+            // Any amount of arguments could be provided
+            ParameterCount::Variable => true,
+        }
+    }
+
+    pub fn to_message(&self, provided: PassedArgumentCount) -> String {
+        match self {
+            ParameterCount::Fixed(required) => {
+                format!("{} arguments but {} were supplied", required, provided)
```
Kampfkarren, line 145: For these, make sure it reads correctly with 1 argument--i.e. "expected 1 arguments" doesn't read properly.

I suspect we will need this again, so should be fine to just have as a helper function in its own module.

selene-lib/src/rules/mismatched_arg_count.rs
```diff
@@ -0,0 +1,425 @@
...
+            }
+            ParameterCount::Minimum(required) => format!(
+                "atleast {} arguments but {} were supplied",
```
Kampfkarren, line 148:
```suggestion
                "at least {} arguments but {} were supplied",
```

selene-lib/src/rules/mismatched_arg_count.rs
```diff
@@ -0,0 +1,425 @@
...
+            ParameterCount::Variable => "a variable amount of arguments".to_owned(),
+        }
+    }
+}
+impl Display for ParameterCount {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            ParameterCount::Fixed(required) => write!(f, "expected {} arguments", required),
+            ParameterCount::Minimum(required) => {
+                write!(f, "expected atleast {} arguments", required)
```
Kampfkarren, line 160:
```suggestion
                write!(f, "expected at least {} arguments", required)
```

selene-lib/src/rules/mismatched_arg_count.rs
```diff
@@ -0,0 +1,425 @@
...
+    }
+}
+
+#[derive(Clone, Copy, Debug)]
+enum PassedArgumentCount {
+    /// Passed a fixed amount of arguments, such as foo(a, b, c) or foo(a, call(), c) or foo(a, ..., c)
+    Fixed(usize),
+    /// Passed a variable of arguments - but we know the lower bound: e.g. foo(a, b, call()) or foo(a, b, ...)
+    Variable(usize),
+}
```
Kampfkarren, line 173:
```suggestion
}

```

selene-lib/src/rules/mismatched_arg_count.rs
```diff
@@ -0,0 +1,425 @@
...
+impl Display for PassedArgumentCount {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            PassedArgumentCount::Fixed(amount) => write!(f, "{} arguments", amount),
+            PassedArgumentCount::Variable(amount) => write!(f, "atleast {} arguments", amount),
```
Kampfkarren, line 178:
```suggestion
            PassedArgumentCount::Variable(amount) => write!(f, "at least {} arguments", amount),
```

selene-lib/src/rules/mismatched_arg_count.rs
```diff
@@ -0,0 +1,425 @@
...
+        }
+    }
+}
+
+fn function_call_argument_count(function_args: &ast::FunctionArgs) -> PassedArgumentCount {
+    match function_args {
```
Kampfkarren, line 184: Same here with `_`, as long as it doesn't lint.

selene-lib/src/rules/mismatched_arg_count.rs
```diff
@@ -0,0 +1,425 @@
...
+        ast::FunctionArgs::Parentheses { arguments, .. } => {
+            // We need to be wary of items with side effects, such as function calls or ... being the last argument passed
+            // e.g. foo(a, b, call()) or foo(a, b, ...) - we don't know how many arguments were passed.
+            // However, if the call is NOT the last argument, as per Lua semantics, it is only classed as one argument, and no side effects occur
+            // e.g. foo(a, call(), b) or foo(a, ..., c)
+
+            let mut passed_argument_count = 0;
+
+            for argument in arguments.pairs() {
+                match argument {
```
Kampfkarren, line 194: Same for non-exhaustive.

selene-lib/src/rules/mismatched_arg_count.rs
```diff
@@ -0,0 +1,425 @@
...
+            PassedArgumentCount::Fixed(amount) => write!(f, "{} arguments", amount),
+            PassedArgumentCount::Variable(amount) => write!(f, "atleast {} arguments", amount),
+        }
+    }
+}
+
+fn function_call_argument_count(function_args: &ast::FunctionArgs) -> PassedArgumentCount {
+    match function_args {
+        ast::FunctionArgs::Parentheses { arguments, .. } => {
+            // We need to be wary of items with side effects, such as function calls or ... being the last argument passed
```
Kampfkarren, line 186: `...` isn't marked as having side-effects, and I don't see this in your tests. I suspect this doesn't work the way you think it does.

selene-lib/src/rules/mismatched_arg_count.rs
```diff
@@ -0,0 +1,425 @@
...
+            // However, if the call is NOT the last argument, as per Lua semantics, it is only classed as one argument, and no side effects occur
+            // e.g. foo(a, call(), b) or foo(a, ..., c)
+
+            let mut passed_argument_count = 0;
+
+            for argument in arguments.pairs() {
+                match argument {
+                    ast::punctuated::Pair::Punctuated(_, _) => passed_argument_count += 1,
+                    ast::punctuated::Pair::End(expression) => {
+                        passed_argument_count += 1;
```
Kampfkarren, line 197: Both branches have this, so just do `if let ast:punctuated::Pair::End(expression) = argument` instead.

selene-lib/src/rules/mismatched_arg_count.rs
```diff
@@ -0,0 +1,425 @@
...
+}
+
+/// A visitor used to map a variable to the necessary number of parameters required
+struct MapFunctionDefinitionVisitor<'a> {
+    scope_manager: &'a ScopeManager,
+    definitions: &'a mut HashMap<Id<Variable>, ParameterCount>,
+    /// Blacklisted variables are ones we will ignore completely - this may be because its a reassigned global variable
+    /// so we cannot determine which function body/parameter count matches.
+    blacklisted_variables: HashSet<Id<Variable>>,
+}
```
Kampfkarren, line 219:
```suggestion
}

```

selene-lib/src/rules/mismatched_arg_count.rs
```diff
@@ -0,0 +1,425 @@
...
+impl Visitor<'_> for MapFunctionDefinitionVisitor<'_> {
+    fn visit_local_function(&mut self, function: &ast::LocalFunction<'_>) {
+        let identifier = range(function.name());
+        let variable = self
+            .scope_manager
+            .variables
+            .iter()
+            .find(|v| v.1.identifiers.contains(&identifier));
```
Kampfkarren, line 227:
```suggestion
            .find(|variable| variable.1.identifiers.contains(&identifier));
```

selene-lib/src/rules/mismatched_arg_count.rs
```diff
@@ -0,0 +1,425 @@
...
+        }
+    }
+
+    fn visit_function_declaration(&mut self, function: &ast::FunctionDeclaration<'_>) {
+        let identifier = range(function.name());
+        if let Some((_, reference)) = self
+            .scope_manager
+            .references
+            .iter()
+            .find(|r| r.1.identifier == identifier)
```
Kampfkarren, line 241:
```suggestion
            .find(|reference| reference.1.identifier == identifier)
```

selene-lib/src/rules/mismatched_arg_count.rs
```diff
@@ -0,0 +1,425 @@
...
+            let expression = expressions.next();
+
+            if let Some(ast::Expression::Value { value, .. }) = expression {
+                if let ast::Value::Function((_, function_body)) = &**value {
+                    let identifier = range(name_token);
+                    let variable = self
+                        .scope_manager
+                        .variables
+                        .iter()
+                        .find(|v| v.1.identifiers.contains(&identifier));
```
Kampfkarren, line 275:
```suggestion
                        .find(|variable| variable.1.identifiers.contains(&identifier));
```

selene-lib/src/rules/mismatched_arg_count.rs
```diff
@@ -0,0 +1,425 @@
...
+                }
+            }
+        }
+    }
+
+    fn visit_local_assignment(&mut self, local_assignment: &ast::LocalAssignment) {
+        let mut expressions = local_assignment.expr_list().iter();
+
+        for name_token in local_assignment.name_list() {
+            let expression = expressions.next();
```
Kampfkarren, line 266: Use `zip`.

selene-lib/src/rules/mismatched_arg_count.rs
```diff
@@ -0,0 +1,425 @@
...
+                }
+            }
+        }
+    }
+
+    fn visit_assignment(&mut self, assignment: &ast::Assignment) {
+        let mut expressions = assignment.expr_list().iter();
+
+        for var in assignment.var_list() {
+            let expression = expressions.next();
```
Kampfkarren, line 290: Use `zip`.

selene-lib/src/rules/mismatched_arg_count.rs
```diff
@@ -0,0 +1,425 @@
...
+
+            if let Some(ast::Expression::Value { value, .. }) = expression {
+                if let ast::Value::Function((_, function_body)) = &**value {
+                    let identifier = range(var);
+                    if let Some((_, reference)) = self
+                        .scope_manager
+                        .references
+                        .iter()
+                        .find(|r| r.1.identifier == identifier)
```
Kampfkarren, line 299:
```suggestion
                        .find(|reference| reference.1.identifier == identifier)
```

selene-lib/src/rules/mismatched_arg_count.rs
```diff
@@ -0,0 +1,425 @@
...
+                    {
+                        if let Some(variable) = reference.resolved {
```
Kampfkarren, line 301: Copy-paste, can we avoid this?

selene-lib/src/rules/mismatched_arg_count.rs
```diff
@@ -0,0 +1,425 @@
...
+    scope_manager: ScopeManager,
+    definitions: HashMap<Id<Variable>, ParameterCount>,
+}
+
+impl Visitor<'_> for MismatchedArgCountVisitor {
+    fn visit_function_call(&mut self, call: &ast::FunctionCall) {
+        if_chain::if_chain! {
+            // Check that we're using a named function call, with an anonymous call suffix
+            if let ast::Prefix::Name(name) = call.prefix();
+            if let ast::Suffix::Call(ast::Call::AnonymousCall(args)) = call.iter_suffixes().next().unwrap();
```
Kampfkarren, line 335:
```suggestion
            if let Some(ast::Suffix::Call(ast::Call::AnonymousCall(args))) = call.iter_suffixes().next();
```

I know this shouldn't be possible, but if we can so easily avoid a panic here I think we should.

selene-lib/src/rules/mismatched_arg_count.rs
```diff
@@ -0,0 +1,425 @@
...
+
+            // Find the corresponding function definition
+            let identifier = range(name);
+            if let Some((_, reference)) = self.scope_manager.references.iter().find(|r| r.1.identifier == identifier);
```
Kampfkarren, line 339:
```suggestion
            if let Some((_, reference)) = self.scope_manager.references.iter().find(|reference| reference.1.identifier == identifier);
```

Kampfkarren (changes requested):
Awesome!

selene-lib/src/rules/mismatched_arg_count.rs
JohnnyMorganz: I did try using `zip` here, with something like `let assignment_expressions = assignment.var_list().iter().zip(assignment.expr_list());`, but I had some awkward lifetime errors, so I just used this aswell (which I got from ScopeManager).
I'll try this again though, and see if I can find what went wrong

selene-lib/src/rules/mismatched_arg_count.rs
JohnnyMorganz: This lints with `unreachable_pattern`

selene-lib/src/rules/mismatched_arg_count.rs
JohnnyMorganz: Nevermind, seems to be working fine now - must have done something wrong last time!

selene-lib/src/rules/mismatched_arg_count.rs
JohnnyMorganz: The `to_message` impl function is specific to this lint - should I remove it from the impl and make it a normal function here?

selene-lib/src/rules/mismatched_arg_count.rs
JohnnyMorganz: Technically this actually doesn't really matter anymore since we are only linting for too-many-arguments, rather than too-little-arguments, so this isn't particularly important anymore.
However, I will fix this for completeness

selene-lib/src/rules/mismatched_arg_count.rs
Kampfkarren: Nevermind then.

selene-lib/src/rules/mismatched_arg_count.rs
```diff
@@ -197,6 +194,20 @@ impl Display for PassedArgumentCount {
     }
 }
 
+fn is_vararg(expression: &ast::Expression) -> bool {
+    if let ast::Expression::Value { value, .. } = expression {
+        if let ast::Value::Symbol(token) = &**value {
```
Kampfkarren, line 195: Let's use if_chain here instead.

selene-lib/src/rules/mismatched_arg_count.rs
```diff
@@ -0,0 +1,449 @@
...
+    }
+
+    false
+}
+
+impl PassedArgumentCount {
+    fn from_function_args(function_args: &ast::FunctionArgs) -> Self {
+        match function_args {
+            ast::FunctionArgs::Parentheses { arguments, .. } => {
+                // We need to be wary of items with side effects, such as function calls or ... being the last argument passed
```
Kampfkarren, line 211: Thinking it over, `has_side_effects` is not what you want here. You just want to check if something is a function call. `{ foo() }`, for instance, has `has_side_effects() == true`, but it doesn't make any difference as to whether something changes variable arguments..

selene-lib/src/rules/mismatched_arg_count.rs
```diff
@@ -0,0 +1,449 @@
...
+
+#[derive(Clone, Copy, Debug)]
+enum PassedArgumentCount {
+    /// Passed a fixed amount of arguments, such as foo(a, b, c) or foo(a, call(), c) or foo(a, ..., c)
+    Fixed(usize),
+    /// Passed a variable of arguments - but we know the lower bound: e.g. foo(a, b, call()) or foo(a, b, ...)
+    Variable(usize),
+}
+
+fn is_vararg(expression: &ast::Expression) -> bool {
```
Kampfkarren, line 193: Move this out to ast_util, I suspect it being useful too.

Kampfkarren (changes requested):
Closer!

selene-lib/src/ast_util/mod.rs
```diff
@@ -37,6 +41,34 @@ pub fn first_code<'ast>(ast: &Ast<'ast>) -> Option<(Position, Position)> {
...
+        } else {
+            false
+        }
+    }
+}
+
+pub fn is_function_call(expression: &ast::Expression) -> bool {
+    if_chain::if_chain! {
+        if let ast::Expression::Value { value, .. } = expression;
+        if let ast::Value::FunctionCall(_) = &**value;
```
Kampfkarren, line 62: This isn't enough for an if_chain, the other one was 4 long so it made more sense.

Kampfkarren:
full-moon updated, make sure it compiles.

docs/src/lints/mismatched_arg_count.md
```diff
@@ -0,0 +1,37 @@
+# mismatched_arg_count
+## What it does
+Checks for too many arguments passed to function calls of defined functions
```
Kampfkarren, line 3:
```suggestion
Checks for too many arguments passed to function calls of defined functions.
```

docs/src/lints/mismatched_arg_count.md
```diff
@@ -0,0 +1,37 @@
...
+    foo = function(a, b, c, d)
+        print("b")
+    end
+end
+
+foo(1, 2, 3, 4) --> "a" [mismatched args]
+updateFoo()
+foo(1, 2, 3, 4) --> "b" [no more mismatched args]
+```
+Selene can not tell that `foo` corresponds to a new definition because `updateFoo()` was called in the current context, without actually "running" the program.
```
Kampfkarren, line 37:
```suggestion
selene can not tell that `foo` corresponds to a new definition because `updateFoo()` was called in the current context, without actually "running" the program.
```

Stylized 😎

docs/src/lints/mismatched_arg_count.md
```diff
@@ -0,0 +1,37 @@
...
+    print("a")
+end
+
+function updateFoo()
+    foo = function(a, b, c, d)
+        print("b")
+    end
+end
+
+foo(1, 2, 3, 4) --> "a" [mismatched args]
```
Kampfkarren, line 33:
```suggestion
foo(1, 2, 3, 4) --> "a" [mismatched args, but selene doesn't know]
```

Just to make it ultra-obvious to someone just skimming through code samples.

Kampfkarren:
!selenology

Kampfkarren:
!selenology

Kampfkarren:
I deleted it by mistake, but it was fine.

PR #249, opened 2021-05-19
Add `onNewLine` and `onIdle` run type configurations

### LINKED ISSUE #236: Choose when to lint via extension

ok-nick:
This pr closes #236 by adding the `onNewLine` and `onIdle` run type configurations.

I'm not too sure about how I handled detecting new lines, I don't believe there is a better approach?

### DIFF
--- a/selene-vscode/src/extension.ts
+++ b/selene-vscode/src/extension.ts
@@ -177,13 +177,13 @@ export async function activate(context: vscode.ExtensionContext) {
             case RunType.OnType:
                 return vscode.workspace.onDidChangeTextDocument(event => lint(event.document))
             case RunType.OnNewLine:
-                return vscode.workspace.onDidChangeTextDocument(event => {
-                    for (const content of event.contentChanges) {
-                        // Contrary to removing lines, adding new lines will leave the range at the same value hence the string comparisons
-                        if (!content.range.isSingleLine || content.text == "\n" || content.text == "\r\n") {
-                            return lint(event.document);
-                        }
-                    }
+				return vscode.workspace.onDidChangeTextDocument(event => {
+					// Contrary to removing lines, adding new lines will leave the range at the same value hence the string comparisons
+					if (event.contentChanges.some(content =>
+						!content.range.isSingleLine || content.text == "\n" || content.text == "\r\n"
+					)) {
+						lint(event.document)
+					}
                 })
             case RunType.OnIdle:
                 return vscode.workspace.onDidChangeTextDocument(event => {
--- a/selene-vscode/src/extension.ts
+++ b/selene-vscode/src/extension.ts
@@ -166,14 +166,13 @@ export async function activate(context: vscode.ExtensionContext) {
     }
 
     function listenToChange() {
-		switch (vscode.workspace.getConfiguration("selene").get<RunType>("run")) {
-            case RunType.OnSave: {
+        switch (vscode.workspace.getConfiguration("selene").get<RunType>("run")) {
+            case RunType.OnSave:
                 return vscode.workspace.onDidSaveTextDocument(lint)
-            }
-            case RunType.OnType: {
+            case RunType.OnType:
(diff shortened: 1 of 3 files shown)


Kampfkarren:
If a PR closes an issue, put "Closes #blablabla" so it is done automatically.

selene-vscode/src/extension.ts
```diff
@@ -1,6 +1,7 @@
 import * as selene from "./selene"
 import * as util from "./util"
 import * as vscode from "vscode"
+import * as timers from "timers"
```
Kampfkarren, line 4: ABC order.

selene-vscode/src/extension.ts
```diff
@@ -165,19 +168,34 @@ export async function activate(context: vscode.ExtensionContext) {
         diagnosticsCollection.set(document.uri, diagnostics)
     }
 
+    let lastTimeout: NodeJS.Timeout
     function listenToChange() {
+        const idleDelay = vscode.workspace.getConfiguration("selene").get<number>("idleDelay") as number
```
Kampfkarren, line 173: We only need this on the OnIdle branch, move it there.

selene-vscode/src/extension.ts
```diff
@@ -165,19 +168,34 @@ export async function activate(context: vscode.ExtensionContext) {
...
             case RunType.OnSave:
                 return vscode.workspace.onDidSaveTextDocument(lint)
             case RunType.OnType:
                 return vscode.workspace.onDidChangeTextDocument(event => lint(event.document))
-
+            case RunType.OnNewLine:
+				return vscode.workspace.onDidChangeTextDocument(event => {
+					// Contrary to removing lines, adding new lines will leave the range at the same value hence the string comparisons
+					if (event.contentChanges.some(content =>
+						!content.range.isSingleLine || content.text == "\n" || content.text == "\r\n"
```
Kampfkarren, line 183: `===`

selene-vscode/src/extension.ts
```diff
@@ -165,19 +168,34 @@ export async function activate(context: vscode.ExtensionContext) {
...
+        const idleDelay = vscode.workspace.getConfiguration("selene").get<number>("idleDelay") as number
         switch (vscode.workspace.getConfiguration("selene").get<RunType>("run")) {
             case RunType.OnSave:
                 return vscode.workspace.onDidSaveTextDocument(lint)
             case RunType.OnType:
                 return vscode.workspace.onDidChangeTextDocument(event => lint(event.document))
-
+            case RunType.OnNewLine:
+				return vscode.workspace.onDidChangeTextDocument(event => {
+					// Contrary to removing lines, adding new lines will leave the range at the same value hence the string comparisons
```
Kampfkarren, line 181: I don't understand this, what if we make a new line in the middle of the script?

Kampfkarren (changes requested):

selene-vscode/src/extension.ts
ok-nick: `range.start.line` and `range.end.line` will stay the exact same as a new line is added. While removing lines they are different. `isSingleLine` is a `boolean` to determine if the range is on different lines which still respects what I’ve said above.
If you make a new line in the middle of the script, then the text would either be `\n` or `\r\n`

selene-vscode/src/extension.ts
```diff
@@ -29,6 +30,8 @@ enum Severity {
 enum RunType {
     OnSave = "onSave",
     OnType = "onType",
+    OnNewLine = "onNewLine",
+    OnIdle = "onIdle"
```
Kampfkarren, line 34:
```suggestion
    OnIdle = "onIdle",
```

PR #250, opened 2021-05-19
Add duplicate keys lint

### LINKED ISSUE #100: Error for duplicate keys in object literal
```lua
return {
    Foo = 1,
    Foo = 2,
}
```

...should error.

JohnnyMorganz:
Closes #100


selene-lib/src/rules/duplicate_keys.rs
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    fn severity(&self) -> Severity {
+        Severity::Warning
+    }
+
+    fn rule_type(&self) -> RuleType {
+        RuleType::Complexity
+    }
+}
```
Kampfkarren, line 54:
```suggestion
}

```

selene-lib/src/rules/duplicate_keys.rs
```diff
@@ -0,0 +1,133 @@
...
+struct DuplicateKey {
+    name: String,
+    position: (usize, usize),
+    original_declaration: (usize, usize),
+}
```
Kampfkarren, line 59:
```suggestion
}

```

selene-lib/src/rules/duplicate_keys.rs
```diff
@@ -0,0 +1,133 @@
...
+    duplicates: Vec<DuplicateKey>,
+}
+
+/// Attempts to evaluate an expression key such as `"foobar"` in `["foobar"] = true` to a named identifier, `foobar`.
+/// Only works for string literal expression keys.
+fn key_expression_to_name(expression: &ast::Expression) -> Option<String> {
+    if let ast::Expression::Value { value, .. } = expression {
+        if let ast::Value::String(token) = &**value {
+            if let tokenizer::TokenType::StringLiteral { literal, .. } = token.token().token_type()
+            {
```
Kampfkarren, line 70: Did Rustfmt do this? I wouldn't suspect it would move it to a new line.

selene-lib/src/rules/duplicate_keys.rs
```diff
@@ -0,0 +1,133 @@
...
+        let mut declared_fields = HashMap::new();
+
+        for field in node.fields() {
+            let field_range = range(field);
+
+            match field {
+                ast::Field::NameKey { key, .. } => {
+                    let key = key.token().to_string();
+                    self.check_field(&mut declared_fields, key, field_range);
+                }
```
Kampfkarren, line 109:
```suggestion
                }

```

selene-lib/src/rules/duplicate_keys.rs
```diff
@@ -0,0 +1,133 @@
...
+                ast::Field::ExpressionKey { key, .. } => {
+                    if let Some(key) = key_expression_to_name(key) {
+                        self.check_field(&mut declared_fields, key, field_range);
+                    }
+                }
```
Kampfkarren, line 114:
```suggestion
                }

```

Kampfkarren (changes requested):
I'm assuming this only works for names, which is fine.

Can you add a test to make sure it doesn't lint with other stuff?

selene-lib/src/rules/duplicate_keys.rs
JohnnyMorganz: Yep it did

JohnnyMorganz:
> I'm assuming this only works for names, which is fine.

Yeah, it will do names, but it can do constant strings:
```lua
local x = {
    foo = true,
    ["foo"] = false,
}
```

I was debating whether to also do it for constant number literals? Its a bit more complex, but I wonder how feasible it is. Also, would it handle array/dictionary mixed:
```lua
local x = {
    true,
    false,
    [1] = true
}
```

Kampfkarren:
That would be good, but not necessary. Up to you.

Kampfkarren:
full-moon updated, make sure it compiles.

docs/src/lints/duplicate_keys.md
```diff
@@ -0,0 +1,30 @@
+# duplicate_keys
+## What it does
+Checks for duplicate keys being defined inside of tables.
+
+## Why this is bad
+Tables with a key defined more than once can lead to unexpected behaviour in your program.
```
Kampfkarren, line 6: Let's be more specific.

```suggestion
Tables with a key defined more than once will only use one of the values.
```

Kampfkarren:
!selenology

Kampfkarren:
!selenology

Kampfkarren:
Nice!

selene-lib/src/rules/duplicate_keys.rs
```diff
@@ -150,6 +150,8 @@ impl Visitor<'_> for DuplicateKeysVisitor {
                     };
                     self.check_field(&mut declared_fields, key, field_range)
                 }
+
+                other => panic!("unknown node {:?}", other),
```
Kampfkarren, line 154: We shouldn't panic! The point of the `non_exhaustive` is so we can add new nodes without having to update consumers.

selene-lib/src/rules/duplicate_keys.rs
```diff
@@ -0,0 +1,181 @@
...
+            })
+            .collect()
+    }
+
+    fn severity(&self) -> Severity {
+        Severity::Warning
+    }
+
+    fn rule_type(&self) -> RuleType {
+        RuleType::Complexity
```
Kampfkarren, line 52:
```suggestion
        RuleType::Correctness
```

This is guaranteed to be wrong, especially with the guarantees the lint has.

Kampfkarren (changes requested):
Last pieces.

selene-lib/src/rules/duplicate_keys.rs
JohnnyMorganz: Whoops, forgot to change this

PR #251, opened 2021-05-20
Improve CLI handling of errors

### LINKED ISSUE #76: VSCode extension stops linting if it encounters a symbol not used in Lua outside of a string literal
Adding a symbol that isn't used in Lua (`?`, `$`, `♧`, etc.) to a file outside of a string causes the VSCode extension to stop linting. Try this code snippet:

```lua
local thisShouldLint?
```

### LINKED ISSUE #203: Parser errors but selene exit with code 0
I ran `selene src`, and it threw this error:

```
thread '<unnamed>' panicked at 'called `Result::unwrap()` on an `Err` 
value: ParseIntError { kind: InvalidDigit }', C:\...\.cargo\registry\src\
github.com-1ecc6299db9ec823\selene-lib-0.10.1\src\rules\
bad_string_escape.rs:129:88
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace     
Results:
0 errors
0 warnings
0 parse errors
```

But the exit code was still zero.

## Repro

Create a file with this:
```lua
return '\97b'
```

Run selene on the file and then you can run `echo $?` to print the last exit code.

JohnnyMorganz:
As tokenizer error fields are now accessibly in full_moon, this PR emits a proper codespan for them, rather than emitting a plain error.
As there is now a rich diagnostic, this will also be picked up in the VSCode extension. This fixes #76.

Also, this PR adds proper handling in case a thread panics to ensure that it exits with a non-zero code, meaning it will be picked up in CI. This closes #203
I didn't, however, change the output. It will currently output something like:
```
thread '<unnamed>' panicked at 'called `Result::unwrap()` on an `Err` value: ParseIntError { kind: InvalidDigit }', selene-lib\src\rules\bad_string_escape.rs:129:88
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
Results:
0 errors
0 warnings
0 parse errors
```
but the results at the end kind of indicate that there were no errors. Should the results output be changed? I believe `errors` and `warnings` only concern lint failures, so I didn't know if it should increment one of those, and its not a parse error either.


Kampfkarren:
I think it's obvious no matter what with the output there was a panic.

Kampfkarren (changes requested):
Needs a CHANGELOG.

Kampfkarren:
Your changelog is borked.

Kampfkarren:
Changelog is conflicted again 😅

CHANGELOG.md
```diff
@@ -1,9 +1,11 @@
 # Changelog
 This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [Unreleased]
+## Unreleased
 ### Added
+- Added `debug.getinfo` to the Roblox standard library. [(#260)](https://github.com/Kampfkarren/selene/issues/260)
```
Kampfkarren, line 6: I *just* realized I typo'd this. Oh well.

PR #252, opened 2021-05-21
Fix math on types plus negative ternary test

### LINKED ISSUE #248: Invalid lint error for CFrame.lookAt in particular case
```lua
BPart.CFrame = CFrame.lookAt(0.5 * (PrevPoint + NextPoint), NextPoint)
```
gives
```
use of standard_library function `CFrame.lookAt` is incorrect
```
The temporary fix below was applied:
```lua
BPart.CFrame = CFrame.lookAt((PrevPoint + NextPoint) * 0.5, NextPoint)
```

Kampfkarren:
Fixes #248.


Kampfkarren:
!selenology

PR #258, opened 2021-06-16
Update full-moon to 0.12

Kampfkarren:


Kampfkarren:
!selenology

matthargett:
An interesting crash:
```
thread '<unknown>' has overflowed its stack
fatal runtime error: stack overflow
Abort trap: 6
```

which I reduced to this:
```lua
return function()
	local Packages = script.Parent.Parent.Parent
	local LuauParty = require(Packages.LuauParty)
	local Fob = LuauParty.Fob

	describe('getNextLines', function()
		describe('given expired lines', function()

			describe('no entangled lines', function()
				describe('pending lines with lower priority than expired lines', function()
					local root
					beforeEach(function()
						root = Fob.assign(root, {
						})
					end)
				end)
			end)
		end)
	end)
end
```

something to do with the nesting of the describe() blocks. If I remove one of them, crash goes away.

matthargett:
weird parse error, which only affects a single type out of several in the file:
```lua
export type ComponentInstance<C, H> = {
  currentTiger: Object,
  instance: any,
  prevProps: Object?,
  props: Object,
  impl: FundamentalImpl<C, H>,
  state: Object
}

export type FundamentalImpl<C, H> = {
  displayName: string,
  reconcileHilts: boolean,
}
```

gets this error:
```
122 │   displayName: string,
    │              ^ expected `}` to match `{`
```

Kampfkarren:
The other one is just because debug full-moon is still kinda cruddy, I need to get stacker working on it at some point.

PR #265, opened 2021-06-21
Fix hex string sequences panicing

### LINKED ISSUE #205: Escape sequence makes selene panic
Using an escape sequence like this will make selene panic (but not exit with an error code, as described in #203):

```lua
return '\97b'
```

The reference manual explains the syntax of escape sequences in section `2.1 - Lexical Conventions`: https://www.lua.org/manual/5.1/manual.html

In summary: when using `\` followed by a digit, it takes up to the next three digits and that makes the byte value in decimal (not octal, as some stack overflow misleading answers seem to think)

Kampfkarren:
Fixes #205


Kampfkarren:
Wait, that lint result makes no sense.

PR #275, opened 2021-07-18
Add task library

Nezuo:
This PR adds the new task library APIs to the Roblox std.

I wasn't sure if the convention was to include `required = false` optional for tuple arguments. Any guidance on that would be appreciated.

Something to note, both `task.delay` and `task.wait` actually take any type for their first argument. I don't know what the behavior of using another type is so I am just enforcing a `number` type.

There isn't any documentation on these APIs yet so all type information was found by just testing them since the APIs just released!

### DIFF
--- a/selene/src/roblox/base.toml
+++ b/selene/src/roblox/base.toml
@@ -648,6 +648,41 @@ required = false
 type = "number"
 required = false
 
+[[task.defer.args]]
+type = "function"
+
+[[task.defer.args]]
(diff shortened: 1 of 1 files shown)


Kampfkarren:
oh whoops i forgot to tell you to make a changelog

Nezuo:
oops, I always forget to update the changelog. looks like you already did it, thanks!

PR #277, opened 2021-07-22
Duplicate condition lint

### LINKED ISSUE #273: Port duplicate condition lint from Luau
> ## DuplicateCondition (24)
>
> When checking multiple conditions via `and/or` or `if/elseif`, a copy & paste error may result in checking the same condition redundantly. This almost always indicates a bug, so a warning is emitted when use of a duplicate condition is detected.
>
> ```lua
> assert(self._adorns[normID1] and self._adorns[normID1]) -- Condition has already been checked on column 8
> ```

Kampfkarren:
Closes #273

### DIFF
--- a/selene-lib/src/rules/duplicate_condition.rs
+++ b/selene-lib/src/rules/duplicate_condition.rs
@@ -121,8 +121,6 @@ impl<'a> QmcBoolTransformer<'a> {
             }
 
             ast::Expression::Value { value, .. } => match &**value {
-                ast::Value::Number(_) | ast::Value::String(_) => Some(qmc::Bool::True),
-
                 ast::Value::Symbol(symbol) => match symbol.token().token_type() {
                     tokenizer::TokenType::Symbol { symbol } => match symbol {
                         tokenizer::Symbol::False | tokenizer::Symbol::Nil => Some(qmc::Bool::False),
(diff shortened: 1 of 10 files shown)


selene-lib/src/rules/duplicate_condition.rs
```diff
@@ -0,0 +1,336 @@
...
+        };
+
+        if transformer.layers == 0 {
+            return;
+        }
+
+        let simplified = qmc_bool.simplify();
+
+        if simplified.len() > 1 {
+            unimplemented!("complicated expression: {}", expression);
```
Kampfkarren, line 211: I know absolutely nothing about QMC and am just hoping selenology will find real code that hits this.

Kampfkarren:
Turns out the duplicate condition lint in Luau is very simple, and doesn't capture things like this:

image

If this one ends up working, it'll be much better.

Kampfkarren:
Okay, so some notes about Luau's duplicate condition lint--it's really *really* specific. QMC isn't going to work here because we are operating on Lua and/or semantics and not actual boolean algebra.

These limitations show up in Luau's lint, where if you don't use *exclusively* and/or, it fails:

image

Kampfkarren:
Some more interesting cases:

image

Kampfkarren:
image
VERY literal

Kampfkarren:
Can't figure this out in time.

PR #283, opened 2021-08-04
Improve missing std files error

### LINKED ISSUE #272: Incorrect error message when a sub-configuration file is not found
My selene.toml contained: 

`std = "roblox+testez"`

When I run selene I get this output here on v0.13:
``` 
$ selene src/
`std = "roblox"`, but there is no roblox.toml in this directory. We are automatically generating one for you now!
By the way, you can do this manually in the future if you need to use new Roblox features with: `selene generate-roblox-std`.
ERROR: Even after generating the `roblox` standard library, we couldn't retrieve the standard library: io error: The system cannot find the file specified. (os error 2)
```

The roblox.toml was present and generated correctly. It was actually not able to find the testez.toml. The error message should have specifically said testez and not roblox.

soutenu:
Fixes #272

### DIFF
--- a/selene/src/main.rs
+++ b/selene/src/main.rs
@@ -424,7 +424,7 @@ fn start(matches: opts::Options) {
                     eprintln!("`std = \"{}\"`, but some files could not be found:", config.std);
 
                     for path in missing_files {
-                        eprintln!("  {}", path.display())
+                        eprintln!("  {}", path.display());
                     }
 
                     error!("Could not find all standard library files");
--- a/selene/src/main.rs
+++ b/selene/src/main.rs
@@ -403,60 +403,56 @@ fn start(matches: opts::Options) {
             std::process::exit(1);
         }
 
-        Err(error) => {
+        Err(_) => {
             if cfg!(feature = "roblox") && config.std.split('+').any(|name| name == "roblox") {
-                let missing_files: Vec<PathBuf> = config.std
-                    .split('+')
-                    .filter(|name| *name != "roblox")
-                    .map(|name| format!("{}.toml", name))
-                    .map(|name| PathBuf::from(&name))
-                    .filter(|path| !path.exists())
-                    .collect();
+                let roblox_std = Path::new("roblox.toml");
 
-                if missing_files.is_empty() {
-                    eprint!("`std = \"roblox\"`, but there is no roblox.toml in this directory. ");
+                if !roblox_std.exists() {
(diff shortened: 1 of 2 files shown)


selene/src/main.rs
```diff
@@ -405,12 +405,31 @@ fn start(matches: opts::Options) {
...
+                    eprintln!("We are automatically generating one for you now!");
+
+                    eprint!("By the way, you can do this manually in the future if you need ");
+                    eprint!("to use new Roblox features with: ");
+                    eprintln!("`selene generate-roblox-std`.");
+                } else {
+                    eprintln!("`std = \"{}\"`, but some files could not be found:", config.std);
+
+                    for path in missing_files {
+                        eprintln!("  {}", path.display())
```
Kampfkarren, line 427:
```suggestion
                        eprintln!("  {}", path.display());
```

Kampfkarren (changes requested):
This is the wrong place to put this check.

This is only going to work:

1. If you have the roblox feature flag enabled
2. If you have roblox in your std

This should be expanded past this.

selene/src/main.rs
```diff
@@ -403,42 +403,57 @@ fn start(matches: opts::Options) {
...
-                    Ok(_) => {
-                        match StandardLibrary::from_config_name(&config.std, Some(&current_dir)) {
-                            Ok(Some(library)) => library,
-
-                            // This is technically reachable if you edit your config while it is generating.
-                            Ok(None) => {
-                                error!("Standard library was empty after generating roblox standard library, did you edit your config while running selene?");
-                                std::process::exit(1);
-                            }
+                let roblox_std = Path::new("roblox.toml");
```
Kampfkarren, line 408: Do you need this variable? Just do `Path::new("roblox.toml").exists()`.

Kampfkarren (changes requested):
One last nitpick in the review.

I am a bit iffy about throwing away the `Err`, but then I noticed we just always assumed it was because the file wasn't found anyway, so I don't think this is introducing any bugs that didn't already exist, so no worries.

Kampfkarren:
Oh shoot, I completely forgot about this PR.

Can you make a changelog, then should be good to go?

soutenu:
Closing this to prevent dirtying the commit history.

Kampfkarren:
Not sure what you mean by that, I squash commit everything--I've just been busy.

PR #291, opened 2021-09-02
Fixes #288

### LINKED ISSUE #288: -q option says --display-style is specified when it isn't
https://github.com/Dionysusnu/Fusion/runs/3476107673
running the command `selene -q .` errors with `error: The argument '--quiet' cannot be used with '--display-style <display-style>'`, even though --display-style is _not_ specified.

soutenu:


selene/src/main.rs
```diff
@@ -111,13 +111,18 @@ fn emit_codespan(
...
 
-    if opts.display_style == opts::DisplayStyle::Json {
-        writeln!(
-            writer,
-            "{}",
-            json_output::diagnostic_to_json(diagnostic).unwrap()
-        )
-        .unwrap();
+    if let Some(display_style) = opts.display_style {
+        if let opts::DisplayStyle::Json = display_style {
```
Kampfkarren, line 115: Just check `if let Some(opts::DisplayStyle::Json) = opts.display_style`, rather than two if-checks with duplicated elses.

Kampfkarren:
Sorry for letting this sit so long, been busy.

PR #299, opened 2021-09-29
remove title-only markdown in document

hituzi-no-sippo:


docs/book.toml
```diff
@@ -4,7 +4,3 @@ language = "en"
 multilingual = false
 src = "src"
 title = "selene Documentation"
-
-[output.html.redirect]
```
Kampfkarren, line 8: You should still keep these so old links people have still work. Changing the rest of the documentation is fine, though.

Kampfkarren (changes requested):

docs/book.toml
hituzi-no-sippo: Thanks for the review.

I think to old links are not there anymore.
This is because the following `git grep` does not show any.

```shell
# This pull request branch
> git grep 'cli/config'

> git grep 'cli/std'

# main branch
> git grep 'cli/config'
docs/book.toml:"/cli/configuration.html" = "/usage/configuration.html"
docs/src/luacheck.md:- You can configure what lints are allowed in the [configuration](./cli/configuration.md#changing-the-severity-of-lints).

> git grep 'cli/std'
docs/book.toml:"/cli/std.html" = "/usage/std.html"
docs/src/contributing.md:Optionally, you can add a `.std.toml` with the same name as the test next to the lua file, where you can specify a custom [stand
ard library](./cli/std.html) to use. If you do not, the Lua 5.1 standard library will be used.
docs/src/lints/incorrect_standard_library_use.md:Checks for correct use of [the standard library](../cli/std.md).
docs/src/lints/undefined_variable.md:If you are using a different standard library where a global variable is defined that selene isn't picking up on, cr
eate a [standard library](../cli/std.md) that specifies it.
docs/src/luacheck.md:- Do you have a custom standard library (custom globals, functions, etc)? Read the [standard library guide](./cli/std.md).
```

Kampfkarren: I meant that people might already have bookmarked or in their browser history.

hituzi-no-sippo: Ohh I see.
Thanks for letting me know.
I will fix it.

hituzi-no-sippo: I fixed the redirect config in a08a760.

Kampfkarren:
Thanks! I hope deleting those files keeps the redirects, though if it doesn't I can fix it.

PR #302, opened 2021-10-01
Fix cargo clippy && fmt

zhiburt:
For some reason cargo fmt changes showed like they changed entire file where's in reality is 2 lines....
It showed better in vscode diff.

### DIFF
--- a/selene-lib/src/ast_util/visit_nodes.rs
+++ b/selene-lib/src/ast_util/visit_nodes.rs
@@ -46,6 +46,7 @@ macro_rules! make_node_visitor {
                 )+
             }
 
+            #[allow(clippy::enum_variant_names)]
             #[derive(Clone, Copy, Debug, Hash, PartialEq, Eq)]
             pub enum VisitorType {
                 $(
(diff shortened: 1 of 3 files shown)


Kampfkarren:
> For some reason cargo fmt changes showed like they changed entire file where's in reality is 2 lines....

It's the newlines, can you try to preserve them if possible?

selene-lib/src/ast_util/visit_nodes.rs
```diff
@@ -46,6 +46,7 @@ macro_rules! make_node_visitor {
                 )+
             }
 
+            #[allow(clippy::enum_variant_names)]
```
Kampfkarren, line 49: What's this for?

Kampfkarren: Oh, because they're camel case...I wonder why I did that.

Kampfkarren: Oh, I see! Fair enough.

image

PR #335, opened 2022-01-06
Add gcinfo function

boatbomber:
Adds the `gcinfo()` function global. It is not deprecated in Roblox since Luau gutted the would-be superseding `collectgarbage()` function due to [sandboxing](https://luau-lang.org/sandbox).


Kampfkarren:
Cool, update the CHANGELOG as well.

Kampfkarren:
Lint failure is mine.

PR #336, opened 2022-01-10
vscode: Path settings to be per-machine scope

mkargus:
The current scope of the path in the VSCode extension doesn't play too nicely when using the sync feature on across different OSes.

Example would be using selene via Foreman on Windows, then set the path settings to use the Windows path would result in breaking the extension when using MacOS or Linux (i.e. Codespaces) with syncing enabled and editing to work on different OSes would result in messing up the Windows machine.

### DIFF
--- a/selene-vscode/package.json
+++ b/selene-vscode/package.json
@@ -41,7 +41,8 @@
                         null
                     ],
                     "default": null,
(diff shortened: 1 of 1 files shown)


Kampfkarren:
I only half understand why this works, but this seems to be what the docs recommend. Thanks!

PR #338, opened 2022-01-25
Fix bit32.replace

nottirb:
`bit32.replace` has 4 parameters, but only 3 have been added to `base.toml` (one is missing).

This PR fixes that by adding the additional parameter to `base.toml`. If I missed anything (this is my first PR) please let me know.

### Current:
(top is RobloxLSP, you can see the selene warning on the bottom)
Current

### Expected:
Expected

### DIFF
--- a/selene/src/roblox/base.toml
+++ b/selene/src/roblox/base.toml
@@ -69,6 +69,9 @@ type = "number"
 
 [[bit32.replace.args]]
 type = "number"
+
+[[bit32.replace.args]]
+type = "number"
 required = false
 
 [[bit32.lrotate.args]]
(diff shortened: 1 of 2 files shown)


selene/src/roblox/base.toml
```diff
@@ -67,6 +67,9 @@ type = "number"
 [[bit32.replace.args]]
 type = "number"
 
+[[bit32.replace.args]]
+type = "number"
```
Kampfkarren, line 71:
```suggestion
type = "number"
required = false
```

nottirb: as far as I could tell the first through third parameter is required:

running
```lua
bit32.replace(1, 1)
```

results in the following error:
```
21:08:29.650  bit32.replace(1, 1):1: missing argument #3 to 'replace' (number expected)  -  Edit
21:08:29.650  Stack Begin  -  Studio
21:08:29.650  Script 'bit32.replace(1, 1)', Line 1  -  Studio
21:08:29.650  Stack End  -  Studio
```

Kampfkarren: Oh, I thought this was width.

What's the 4th parameter in your screenshot, then? Why doesn't your PR include that?

nottirb: The 4th parameter (width) is already accounted for:
Screenshot_42
Screenshot_41

This just seeks to add the third required variable.

Kampfkarren: Gotcha :+1:

PR #344, opened 2022-03-01
Add line/col locations to JSON output

deathlyfrantic:
Hi! I wanted to add Selene support to ALE for Vim/Neovim but it would be a lot easier to do if line/column were reported in the JSON output, so I took a stab at doing it. It's been a while since I've worked with Rust so I'm sure this is a mess and needs a lot of work - this is literally the first thing I got to compile. 😄

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 # Changelog
 This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-Unreleased
+## Unreleased
 ### Added
 - Added `start_line`, `start_column`, `end_line`, and `end_column` to JSON diagnostic output.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,10 @@
 # Changelog
 This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+Unreleased
+### Added
+- Added `start_line`, `start_column`, `end_line`, and `end_column` to JSON diagnostic output.
+
 ## [0.16.0] - 2022-01-30
 ### Added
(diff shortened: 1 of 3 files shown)


selene/src/json_output.rs
```diff
@@ -22,33 +22,44 @@ struct Label {
...
+    start_line: usize,
+    start_column: usize,
     end: usize,
+    end_line: usize,
+    end_column: usize,
 }
 
-fn label_to_serializable<T>(label: &CodespanLabel<T>) -> Label {
+fn label_to_serializable(label: &CodespanLabel<codespan::FileId>, files: &codespan::Files<&str>) -> Label {
+    let start_location = files.location(label.file_id, label.range.start as u32).unwrap();
```
Kampfkarren, line 33: Can you make both of these use `expect` with an error message? It's a lot easier to debug if something goes wrong.

selene/src/json_output.rs
```diff
@@ -22,33 +22,44 @@ struct Label {
...
             end: label.range.end,
+            end_line: end_location.line.into(),
+            end_column: end_location.column.into(),
         },
     }
 }
 
 pub fn diagnostic_to_json(
     diagnostic: &CodespanDiagnostic<codespan::FileId>,
+    files: &codespan::Files<&str>
```
Kampfkarren, line 50:
```suggestion
    files: &codespan::Files<&str>,
```

Are you running `cargo fmt`? 😉

Kampfkarren (changes requested):
Put this in CHANGELOG too please

Kampfkarren:
Name it Unreleased

CHANGELOG.md
```diff
@@ -1,6 +1,10 @@
 # Changelog
 This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+Unreleased
```
Kampfkarren, line 4:
```suggestion
## Unreleased
```

PR #364, opened 2022-06-04
Deprecate TOML standard library format, introduce new YML based format. Roblox standard libraries now generate outside of directories

### LINKED ISSUE #125: Save roblox.toml somewhere not on the project, and update every few days
It's cleaner and makes it so it never shows up in the git repository.

### LINKED ISSUE #363: Standard library format is unsustainable
Ignoring the fact that TOML for this is unreadable (which is easily fixable by just adding another serde frontend like StrictYAML), any additions to the standard library format are not forwards compatible.

I soon will be adding a `deprecated` field to the standard library. This will break any standard libraries that have something like `[tools.deprecated]`. 

The `selene` global is also odd--it is currently impossible to have a global called `selene`, which could prove annoying if *we* want to use it for plugins.

The entire standard library format needs to be completely redone to support fields forwards compatibility. In this same breath, we should also start using StrictYAML or similar.

The old standard library format should be kept around as selene is at the point of notability where I am uncomfortable making large breaking changes like that.

Kampfkarren:
Closes #363 (Standard library is unsustainable)
Closes #125 (Save roblox.toml somewhere outside directory)

TODO:
- [x] Docs
- [x] Changelog
- [x] Put roblox.yml in an off-directory file (then warn for roblox.toml existing)
- [x] TODOs


selene/src/roblox/collect_std.rs
```diff
@@ -0,0 +1,129 @@
...
+    current_directory: &Path,
+) -> color_eyre::Result<StandardLibrary> {
+    let (cached_library, output_directory, output_location) = match config.roblox_std_source {
+        RobloxStdSource::Floating => {
+            let floating_file_directory = floating_file_directory()?;
+            let floating_file = floating_file_directory.join("roblox.yml");
+
+            if floating_file.exists() {
+                let mut library: StandardLibrary = serde_yaml::from_reader(
+                    fs::File::open(&floating_file)
```
LPGhatguy, line 26: You should wrap this in a `BufReader`, which can result in major performance improvements for readers!

PR #367, opened 2022-06-05
Add deprecated functions and fields

### LINKED ISSUE #11: Let standard libraries support deprecated fields
Will give a warning for use, rather than an error.

### LINKED ISSUE #361: Generated roblox.toml incorrectly thinks `Enum.Font.GothamSemibold` does not exist
Recently (as of today) we've discovered that `Enum.Font.GothamSemibold` is no longer recognized by Selene and we are seeing the following error in CI:

Screen Shot 2022-06-01 at 1 10 24 PM

I diffed our older roblox.toml against the recently generated one, and it looks like for some reason that `GothamSemibold` was replaced by `GothamMedium`

Screen Shot 2022-06-01 at 1 09 56 PM

It seems like GothamMedium might supercede GothamSemibold, but both are indexable from within Studio so /shrug

Kampfkarren:
Closes #361
Closes #11 

- [x] Add deprecated `:connect` and stuff
- [x] Remove deprecated parameter from generate-roblox-std
- [x] Document (both lint and the standard library addition)
- [ ] TODOs (for configuration validation)


Kampfkarren:
For posterity: I am choosing not to correctly mark `wait` and friends as deprecated until selene has fixers. Changing this would cause legacy codebases to get hundreds of warnings without any easy way to fix them, meaning people would just disable the lint.

Kampfkarren:
Validation is fairly difficult unless I write a parser replacement for the regex, which I don't wanna right now

PR #380, opened 2022-06-22
Mutability observations

### LINKED ISSUE #353: Expand `unused_variable` to know `table.insert` does not read the value
Right now:

```lua
local x = 1
x = 2
```

...lints.

This should be expanded (probably with a std-TOML mark) to support the following:

```lua
local x = {}
table.insert(x, 1) -- This is not actually reading the value
```

Kampfkarren:
Closes #353

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
+### Added
+- `table.insert(x)` no longer counts as a read to `x`, which allows selene to alert you that you are only assigning to it.
+  - This is done through a new standard library field for arguments called `observes`. This takes 3 values: "read-write" (the default), signifying a potential read and write, "read", signifying only a read, and "write", signifying only a write. Only "write" has meaning at this time.
+
 ### Fixed
 - Fixed a bunch of performance failures, lowering some benchmarks from 3 seconds to 200ms.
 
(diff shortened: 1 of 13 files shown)


Kampfkarren:
The `client_names` and `buf_client_names` thing there is a real failure, but this one isn't so simple:

```
warning[unused_variable]: new_config is assigned a value, but never used

9  │ on_new_config = function(new_config, new_root_dir)
   │                          ^^^^^^^^^^
10 │ if new_root_dir then
11 │ table.insert(new_config.cmd, '-s')
   │ ---------- `table.insert` only writes to `new_config`
12 │ table.insert(new_config.cmd, new_root_dir)
   │ ---------- `table.insert` only writes to `new_config`
```

...because this is from an argument.

Kampfkarren:
The zombie-strike one is also incorrect:

```lua
	local roomTypes = {
		boss = {},
		bossBefore = {},
		enemy = {},
		obby = {},
		treasure = {},
		wave = {},
	}

	for _, room in pairs(roomModels:GetChildren()) do
		local obbyType = room:FindFirstChild("ObbyType")
		if obbyType ~= nil then
			local roomTable = assert(roomTypes[obbyType.Value])
			table.insert(roomTable, room)
		end
	end
```

...but the `ignore` one is properly accurate.

Kampfkarren:
The solution is most likely to only kick in when the variable is a statically declared table. Userdata is going to be impossible to treat properly regardless.

PR #396, opened 2022-07-30
Match .luau filename extension by default

### LINKED ISSUE #325: Support `.luau` as an extension for files when Roblox feature is active
Luau has formally adopted `.luau` as it's extension (see: https://github.com/Roblox/luau/issues/97) so we should support it in Selene too.

Yumacide:
Closes #325.
It seems like the glob crate doesn't allow pattern lists like `**/*.{lua,luau}` (see [docs](https://docs.rs/glob/latest/glob/struct.Pattern.html)), so I chose to match multiple patterns instead. ~~I also renamed `--pattern` to `--patterns` to match the change, let me know if that's not okay.~~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,13 @@
 # Changelog
 This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Changed
+
+- Match `.luau` filename extension by default.
+- Allow multiple comma-separated glob patterns with `--patterns` rather than `--pattern`.
+
 ## [0.20.0] - 2022-07-21
 ### Added
 - Added [`utf8` globals](https://q-syshelp.qsc.com/Content/Control_Scripting/Lua_5.3_Reference_Manual/Standard_Libraries/4_-_Basic_UTF-8_Support.htm) to the builtin `lua53` standard library.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 ### Changed
 
 - Match `.luau` filename extension by default.
-- Allow multiple comma-separated glob patterns with `--patterns` rather than `--pattern`.
+- Allow multiple comma-separated glob patterns with `--pattern`.
 
 ## [0.20.0] - 2022-07-21
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 ### Changed
 
 - Match `.luau` filename extension by default.
-- Allow multiple comma-separated glob patterns with `--pattern`.
+- Allow `--pattern` to be passed multiple times.
 
 ## [0.20.0] - 2022-07-21
 ### Added
--- a/docs/src/cli/usage.md
+++ b/docs/src/cli/usage.md
@@ -1,5 +1,4 @@
 # CLI Usage
-
 If you want to get a quick understanding of the interface, simply type `selene --help`.
 
 ```
@@ -19,7 +18,7 @@ OPTIONS:
         --display-style <display-style>    Sets the display method [possible values: Json, Rich, Quiet]
         --num-threads <num-threads>        Number of threads to run on, default to the numbers of logical cores on your
                                            system [default: your system's cores]
-        --patterns <patterns>              Comma-separated globs to match files to check [default: **/*.lua,**/*.luau]
+        --pattern <pattern>              Comma-separated globs to match files to check [default: **/*.lua,**/*.luau]
 
 ARGS:
     <files>...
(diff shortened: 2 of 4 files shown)


docs/src/cli/usage.md
```diff
@@ -72,10 +73,10 @@ Results:
 0 parse errors
 ```
 
-**--num-threads** *num-threads*
+**--num-threads** _num-threads_
```
Kampfkarren, line 76: Revert the unrelated markdown changes (this and some whitespace stuff)

CHANGELOG.md
```diff
@@ -1,6 +1,13 @@
 # Changelog
 This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Changed
+
+- Match `.luau` filename extension by default.
+- Allow multiple comma-separated glob patterns with `--patterns` rather than `--pattern`.
```
Kampfkarren, line 9: `--pattern` should just take multiple patterns, I don't want to break backwards compatibility with the older parameter, and commas already aren't allowed in filenames in any of the operating systems we support, from what I know

docs/src/cli/usage.md
```diff
@@ -72,10 +73,10 @@ Results:
...
 
 Specifies the number of threads for selene to use. Defaults to however many cores your CPU has. If you type `selene --help`, you can see this number because it will show as the default for you.
 
-**--pattern** *pattern*
+**--patterns** _patterns_
 
-A [glob](https://en.wikipedia.org/wiki/Glob_(programming)) to match what files selene should check for. For example, if you only wanted to check files that end with `.spec.lua`, you would input `--pattern **/*.spec.lua`. Defaults to `**/*.lua`, meaning "any lua file".
+Comma-separated [globs](<https://en.wikipedia.org/wiki/Glob_(programming)>) to match what files selene should check for. For example, if you only wanted to check files that end with `.spec.lua`, you would input `--patterns **/*.spec.lua`. Defaults to `**/*.lua,**/*.luau`, meaning "any lua/luau file".
```
Kampfkarren, line 82: Why the `<>`?

selene/src/opts.rs
```diff
@@ -7,9 +7,9 @@ use structopt::{clap::arg_enum, StructOpt};
 #[structopt(setting(structopt::clap::AppSettings::ArgsNegateSubcommands))]
 #[structopt(setting(structopt::clap::AppSettings::SubcommandsNegateReqs))]
 pub struct Options {
-    /// A glob to match files with to check
-    #[structopt(long, default_value = "**/*.lua")]
-    pub pattern: String,
+    /// Comma-separated globs to match files to check
+    #[structopt(long, default_value = "**/*.lua,**/*.luau")]
```
Kampfkarren, line 11: We probably shouldn't default match `.luau` unless the "roblox" feature flag is enabled (which will eventually be renamed to Luau).

You can do this with something like:

```suggestion
    #[cfg_attr(feature = "roblox", structopt(long, default_value = "**/*.lua,**/*.luau"))]
    #[cfg_attr(not(feature = "roblox"), structopt(long, default_value = "**/*.lua,**/*.luau"))]
```

Kampfkarren (changes requested):
Neat!

docs/src/cli/usage.md
Yumacide: This and the whitespace changes were done automatically by my editor

CHANGELOG.md
Yumacide: How would I do that?

Kampfkarren: Just don't change the name :P

Yumacide: Oh, I thought you were suggesting something like `selene --pattern foo --pattern foo`.

Yumacide:
Ready to merge! In retrospect, this would be best as an upstream change to `glob`, but I can't find any discussion on it there.

Kampfkarren:
I think glob already supports this with some weird syntax with pipelines, it's not as intuitive

Yumacide:
If that's the case, it's not documented: https://docs.rs/glob/latest/glob/struct.Pattern.html

Yumacide:
Turns out commas are allowed in filenames on windows & *nix -- maybe it's a better idea to do multiple pattern args like `selene --pattern "**/*.lua" --pattern "**/*.luau"` after all.

Kampfkarren:
Yeah that'll work

selene/src/opts.rs
```diff
@@ -1,15 +1,34 @@
...
+use std::{ffi::OsString, path::PathBuf, str::FromStr};
 
 use structopt::{clap::arg_enum, StructOpt};
 
+#[derive(Clone, Debug)]
+pub struct Patterns {
+    pub vec: Vec<String>,
+}
+
+impl FromStr for Patterns {
```
Yumacide, line 10: I had to do this because structopt's `default_value` has to be a string literal; I can't just make it a `Vec<String>` like it should normally be.

selene/src/opts.rs
```diff
@@ -1,15 +1,34 @@
...
 #[structopt(rename_all = "kebab-case")]
 #[structopt(setting(structopt::clap::AppSettings::ArgsNegateSubcommands))]
 #[structopt(setting(structopt::clap::AppSettings::SubcommandsNegateReqs))]
 pub struct Options {
     /// A glob to match files with to check
-    #[structopt(long, default_value = "**/*.lua")]
-    pub pattern: String,
+    #[cfg_attr(
+        feature = "roblox",
+        structopt(long, default_value = "**/*.lua\0**/*.luau")
```
Yumacide, line 28: It seems like `\0` and `/` are the only universally prohibited characters in filenames

Yumacide: This is just for the default value, though. Users would just pass `--pattern` multiple times

selene/src/main.rs
```diff
@@ -347,9 +347,15 @@ fn read_file(checker: &Checker<toml::value::Value>, filename: &Path) {
     );
 }
 
-fn start(matches: opts::Options) {
+fn start(mut matches: opts::Options) {
     *OPTIONS.write().unwrap() = Some(matches.clone());
 
+    if matches.pattern.is_empty() {
```
Kampfkarren, line 353: FWIW this is what I was going to suggest 😄

PR #400, opened 2022-08-10
Change `severity` and `rule_type` to constants

Yumacide:


selene-lib/src/rules.rs
```diff
@@ -55,8 +55,8 @@ pub trait Rule {
         ast_context: &AstContext,
     ) -> Vec<Diagnostic>;
 
-    fn severity(&self) -> Severity;
-    fn rule_type(&self) -> RuleType;
+    fn severity() -> Severity;
```
Kampfkarren, line 58: Any chance you can just make these `const SEVERITY: Severity` etc? If not these should be const fn

Yumacide: Sure! That'll be an annoying regex to write though

Kampfkarren:
Thanks!

PR #401, opened 2022-08-13
Add changelog links

LastTalon:
This improves documentation slightly by implementing links to releases for each version in the changelog.

Some of the versions link to other notes because they were released with their notes on a subsequent release.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,14 +1,14 @@
 # Changelog
 This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [Unreleased]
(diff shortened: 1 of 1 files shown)


Kampfkarren:
Can you have the links for these be in the headers themselves? After that, I might have to adjust the shell script for extracting changelogs.

LastTalon:
Sure, I can do that. Typically I would recommend grouping them, since updating them and changing them is easier with them together.

Is there something the script does that it should be in the headers?

Kampfkarren:
The script is here: https://github.com/Kampfkarren/selene/blob/main/misc/extract-changelog.sh

I think right now it just uses the most recent header, so should be okay.

LastTalon:
It appears to do the same thing after the changes. You may want to check yourself, since I'm unsure exactly what its supposed to look like.

PR #402, opened 2022-08-26
Fixes three separate lints incorrectly handling overriding functions 

### LINKED ISSUE #157: Defining a global and attempting to use it in a function body gives an unused variable warning
image
image
image
image

Kampfkarren:
The following code:

```lua
function global.name() end
```

...would cause 3 separate bugs:

1. `undefined_variable` would not bother checking for `global` (inadequate)
2. `unscoped_variable` would see the write to `global` without a definition, and lint (incorrect)
3. `unused_variable` would tag `global` as unused if `global` was in the standard library (incorrect)

This should help out non-Roblox codebases where code like this isn't uncommon, such as in LOVE.

Fixes #157.


Kampfkarren:
!selenology

PR #405, opened 2022-08-27
Add `--allow-warnings` option to suppress errors on only warnings

### LINKED ISSUE #404: Non-zero exit code on only lint warnings
Currently, Selene exits the process with code 1 if there are only lint warnings. I expect Selene to, like other linters (ESLint), exit with an exit code of 0 here and only exit with non-zero exit codes on lint errors/parse errors.

image

guidojw:
Resolves #404


Kampfkarren:
As per https://github.com/Kampfkarren/selene/issues/404#issuecomment-1229344395, can you make this an opt in `--allow-warnings` flag?

Issue #222, opened 2021-03-27
Custom lints with Lua


Kampfkarren:
I want to be able to write custom lints with Lua, store them in some directory, and have Selene run those. These lints would be similar to the custom JS lints ESLint supports. 

Every script would represent one lint, and return a table just like the `Lint` trait, with information such as its severity, rule type, and most importantly a function that takes in an Ast and returns a table of diagnostics.

These should not run by default for security purposes. Somehow, you must give Selene permission to run them (the extension would just show a basic prompt), where Selene would then remember to trust that directory in a folder outside of the repository, so people cannot simply commit a "yes, trust this" file. I am not excellent at this sort of stuff, so some more work may need to be done, such as allowing different permission levels to reset if any lints change hashes.

Ideally, you might even be able to download repositories of lints from the web, meaning people can create specialized/opinionated linting suites, such as StandardJS.

I also want to allow fixers inside these custom lints as well, and ideally in some of the standard selene lints too. This is because if someone sees a thousand lints from something style related, they'll simply turn it off instead of fixing it. This may be out of scope for this issue though.

These lints would also be under a fixed namespace, such as `plugin::` or similar. This is to prevent name collisions with selene lints, especially in the case of a Lua lint becoming so popular that it deserves promotion to the upstream.

This relies on Lua bindings for full-moon, something I will not start work on until 1.0.0 finalizes the types to some degree, which should be coming soon.

Kampfkarren:
We should probably take a permissions route like Deno does, limiting things like `io` and `require`'ing C libraries. @LPGhatguy considered https://github.com/bazelbuild/starlark for Remodel and Rojo configuration, I might want to look and see if that's a reasonable path for us as well, though I do think I want the ability to at least *allow* io interactions.

Kampfkarren:
full-moon keeps making breaking changes as Luau gets more and more complex, in ways that we believe are impossible to protect ourselves from ahead of time (before an RFC is made). This poses a very large problem for stability of lints.

@JohnnyMorganz suggests stabilizing selene lints with just exposing Lua 5.1 types, as we haven't been breaking backwards compatibility on those. There are *some* changes I want to make, but they're only on the Rust side (renaming `TokenReference` and such). I think that's a good idea.

In the future, I think it *could* be possible to package every *major* full-moon version (especially once it actually does hit 1.0.0), and let the user of a plugin decide what full-moon version they want. Something like either specifying it in a config, or if these are single Lua files, then something like:

```lua
[...]

PR #415, opened 2022-09-15
[EXTREME WIP] Lua plugins for custom lints

Kampfkarren:
So WIP I don't even want to write a body right now

#222


CyanoTex:
Been a while there. Still working on it?

PR #420, opened 2022-09-17
Remove rbx_reflection, various roblox_incorrect_roact_usage buffs

### LINKED ISSUE #59: Check Roact.Event.XXXXX in roblox_incorrect_roact_usage
Verify event is one that is valid for the class.

### LINKED ISSUE #297: Invalid Roact usage reports "Instance" rather than the real class name
```lua
Gradient = e("UIGradient", {
	Angle = math.atan(difference / LINE_HEIGHT),
})
```

...tells you:

> `Angle` is not a property of `Instance`

...rather than "UIGradient".

Kampfkarren:
Fixes #297.
Closes #59.

Roblox standard libraries are now guaranteed to regenerate when the previously generated standard library is on a different version.


Kampfkarren:
I can't reproduce that selenology output, I think it is only creating the old roblox.yml?

Kampfkarren:
Yeah, it's just using cached roblox.yml. Changed to delete every time.

Kampfkarren:
Ugh. I'll deal with it tomorrow.

Kampfkarren:
Actually that doesn't even make sense considering this version would be regenerating it no matter what...

Kampfkarren:
x_x swapped

Kampfkarren:
I keep finding bugs and none of them are the right one!

Kampfkarren:
I fixed the bug in zombie-strike, then updated selenology to also clear old roblox.toml files (but it was a good thing I didn't before this!)

PR #424, opened 2022-09-18
Add `--exclude-dir` option

kyojindev:
Implements an `--exclude-dir` option which allows omitting a directory from being linted. Useful for those who use external libraries in their projects.

### DIFF
--- a/selene/src/main.rs
+++ b/selene/src/main.rs
@@ -479,6 +479,8 @@ fn start(mut matches: opts::Options) {
 
     let pool = ThreadPool::new(matches.num_threads);
 
+    let exclude_path = &matches.exclude_path.as_ref();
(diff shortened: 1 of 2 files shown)


Kampfkarren:
This shouldn't be a command line argument, and especially not one that can only be used once. This should instead be a glob in selene.toml, like:

```toml
std = "roblox"
exclude = ["vendor/"]
```

kyojindev:
> This shouldn't be a command line argument, and especially not one that can only be used once. This should instead be a glob in selene.toml, like:
> 
> ```toml
> std = "roblox"
> exclude = ["vendor/", "*.spec.lua"]
> ```

I agree, I should have thought it out more before making a PR.

PR #425, opened 2022-09-18
Add exclude option in selene.toml

kyojindev:
This implements an `exclude` glob option in selene.toml which excludes any files matching with the globs from being linted.

### DIFF
--- a/selene/Cargo.toml
+++ b/selene/Cargo.toml
@@ -20,6 +20,7 @@ color-eyre = "0.6.1"
 dirs = "4.0.0"
 full_moon = "0.15.1"
 glob = "0.3"
+globset = "0.4.9"
 lazy_static = "1.4"
 num_cpus = "1.10"
 profiling = { version = "1.0.6" }
--- a/selene/src/main.rs
+++ b/selene/src/main.rs
@@ -471,10 +471,10 @@ fn start(mut matches: opts::Options) {
 
     let mut builder = globset::GlobSetBuilder::new();
     for pattern in &config.exclude {
-        builder.add(match globset::Glob::new(&format!("{}", pattern)) {
+        builder.add(match globset::Glob::new(&pattern) {
             Ok(glob) => glob,
             Err(error) => {
-                error!("Invalid glob pattern: {}", error);
+                error!("Invalid glob pattern: {error}");
                 return;
             }
         });
@@ -532,6 +532,7 @@ fn start(mut matches: opts::Options) {
                                     if exclude_set.is_match(&path) {
                                         continue;
                                     }
+
                                     let checker = Arc::clone(&checker);
 
                                     pool.execute(move || read_file(&checker, &path));
--- a/selene/src/main.rs
+++ b/selene/src/main.rs
@@ -469,6 +469,25 @@ fn start(mut matches: opts::Options) {
             }
         };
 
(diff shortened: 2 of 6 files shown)


selene/Cargo.toml
```diff
@@ -20,6 +20,7 @@ color-eyre = "0.6.1"
 dirs = "4.0.0"
 full_moon = "0.15.1"
 glob = "0.3"
+globset = "0.4.9"
```
Kampfkarren, line 23: Can you replace our one usage of glob if you're going to add globset?

selene/src/main.rs
```diff
@@ -469,6 +469,25 @@ fn start(mut matches: opts::Options) {
             }
         };
 
+    let mut builder = globset::GlobSetBuilder::new();
+    for pattern in &config.exclude {
+        builder.add(match globset::Glob::new(&format!("{}", pattern)) {
+            Ok(glob) => glob,
+            Err(error) => {
+                error!("Invalid glob pattern: {}", error);
```
Kampfkarren, line 477:
```suggestion
                error!("Invalid glob pattern: {error}");
```

selene/src/main.rs
```diff
@@ -469,6 +469,25 @@ fn start(mut matches: opts::Options) {
             }
         };
 
+    let mut builder = globset::GlobSetBuilder::new();
+    for pattern in &config.exclude {
+        builder.add(match globset::Glob::new(&format!("{}", pattern)) {
```
Kampfkarren, line 474:
```suggestion
        builder.add(match globset::Glob::new(&pattern) {
```

selene/src/main.rs
```diff
@@ -510,6 +529,9 @@ fn start(mut matches: opts::Options) {
                         for entry in glob {
                             match entry {
                                 Ok(path) => {
+                                    if exclude_set.is_match(&path) {
+                                        continue;
+                                    }
```
Kampfkarren, line 534:
```suggestion
                                    }

```

Kampfkarren (changes requested):
Add it to CHANGELOG.md and to the documentation for selene.toml as well

selene/Cargo.toml
kyojindev: globset doesn't currently support iterating recursively through matching file paths, so I'd have to rewrite that entire section and do it manually. I can do it but I want to get the go-ahead from you before spending time on this.

Kampfkarren: Hmm, so does that mean something like `a/**/b.lua` isn't possible to exclude? Or `a/**/*.spec.lua`? If glob behavior is different between glob and globset then we should just be using glob, and running over every glob itself. Maybe we should just be doing that anyway, I'm unsure what benefit globset has to just looping over a vector of globs.

kyojindev: Sorry, I might have worded this wrong. Globset does allow those exclusions - what it doesn't allow is using a glob as an iterator like we currently do to recursively get the files to lint when provided with a directory.

Kampfkarren: Works for me :+1:

Kampfkarren:
Conflicts with master now

Kampfkarren:
@OverHash Yeah, didn't catch it

PR #439, opened 2022-10-03
Use Cargo workspace fields

OverHash:
Condenses `selene` and `selene-lib` to use the `Cargo.toml` at the root of the repository rather than requiring to update each Cargo file individually.

Note that one downside is that this requires Cargo 1.64 as a MSRV. See the [release notes](https://blog.rust-lang.org/2022/09/22/Rust-1.64.0.html#cargo-improvements-workspace-inheritance-and-multi-target-builds) from Rust 1.64 for more details.

Some additional points of discussion
- Should authors, categories, and keywords be in the workspace manifest? I felt as though these fields might differ between `selene` and `selene-lib`.
- `selene` has a `documentation` field, while `selene-lib` does not. Is this intentional?
  - note that setting the `documentation` field will not allow the default `docs.rs` link to appear.
- We could also specify dependencies at the workspace level.


Cargo.toml
```diff
@@ -4,3 +4,11 @@ members = [
...
     "selene-lib",
 ]
+
+[workspace.package]
+version = "0.21.1"
+edition = "2021"
+homepage = "https://kampfkarren.github.io/selene/"
+license = "MPL-2.0"
+repository = "https://github.com/Kampfkarren/selene"
+rust-version = "1.64"
```
Kampfkarren, line 14: Don't include this please, I never want to commit to an MSRV.

Kampfkarren:
> Should authors, categories, and keywords be in the workspace manifest? I felt as though these fields might differ between selene and selene-lib.

Yes

> selene has a documentation field, while selene-lib does not. Is this intentional?
>    note that setting the documentation field will not allow the default docs.rs link to appear.

Yes, selene-lib is pretty much just internal, and selene has no open APIs you'd want to use.

full-moon should be set as a dependency at the workspace, I set it like that in the plugins branch

Cargo.toml
Kampfkarren: selene is basically never guaranteed to compile on anything other than latest because I always like using new features that come out, and so I'd always be needing to update this, making it fairly useless for anyone trying to actually rely on it.

OverHash: Fair enough, removed!

Cargo.toml
```diff
@@ -7,7 +7,14 @@ members = [
 
 [workspace.package]
 version = "0.21.1"
+authors = ["Kampfkarren <kampfkarren@gmail.com>"]
+categories = ["command-line-utilities"]
+keywords = ["lua", "linter", "lua51"]
```
Kampfkarren, line 12: Don't include categories and keywords, actually. selene-lib isn't really meant to be advertised.

Kampfkarren:
Thanks!

Issue #408, opened 2022-08-28
Standard library loading doesn't support `yaml` extension


LastTalon:
When loading libraries configured in `selene.toml` such as `std = "roblox+testez"`, selene fails to load `testez.yaml` when its present. Works as intended after changing the file to `testez.yml`, but both of these are valid extensions for yaml files.

sisco0:
Could you kindly ensure that the given patch in PR #441 solves the issue as expected @LastTalon @Kampfkarren?

Thank you in advance!

PR #441, opened 2022-10-03
Add YAML as a valid extension for the settings file

sisco0:
The current PR solves #408 by appending `yaml` as the list of valid extensions for the `std` configuration files.

The current solution **requires to be tested** by the issue creator or a project maintainer.

Also, considering that this is my first contribution to the project, could you kindly set a review on my proceeding @Kampfkarren?

Thank you in advance.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - Added `--allow-warnings` option to have selene pass when only warnings occur.
 - Added the ability to allow specific patterns in the `deprecated` lint.
 - Added exclude option to selene.toml for excluding files from lints
-- Added `testez.yaml` as a supported _Testez_ configuration file.
+- Adds support for `.yaml` extensions to be used for standard libraries alongside `.yml`.
 
 ### Changed
 - Updated internal parser, giving substantial parsing speed increases.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - Added `--allow-warnings` option to have selene pass when only warnings occur.
 - Added the ability to allow specific patterns in the `deprecated` lint.
 - Added exclude option to selene.toml for excluding files from lints
+- Added `testez.yaml` as a supported _Testez_ configuration file.
 
 ### Changed
 - Updated internal parser, giving substantial parsing speed increases.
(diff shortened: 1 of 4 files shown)


docs/src/roblox.md
```diff
@@ -24,7 +24,7 @@ Roblox has provided an open source testing utility called [TestEZ](https://roblo
 
 `std = "roblox+testez"`
 
-But first you'll need to create a `testez.yml` file, which you can do so [with this template](https://gist.github.com/Kampfkarren/f2dddc2ebfa4e0662e44b8702e519c2d).
+But first you'll need to create a `testez.yml` or `testez.yaml` file, which you can do so [with this template](https://gist.github.com/Kampfkarren/f2dddc2ebfa4e0662e44b8702e519c2d).
```
Kampfkarren, line 27: Let's just recommend `testez.yml` for now, `.yaml` being supported is something we can put in a separate documentation file, the standard library one.

Kampfkarren (changes requested):
You also need to support this in https://github.com/Kampfkarren/selene/blob/462456995a8efbc9b0b009db2293dcfd8d2bd8f6/selene/src/standard_library.rs#L74 to check for .yml or .yaml.

Also make sure to add this to the `Added` section of the CHANGELOG.

Thanks for the PR!

docs/src/roblox.md
sisco0: The change has been reverted and the file is now untouched.

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - Added `--allow-warnings` option to have selene pass when only warnings occur.
 - Added the ability to allow specific patterns in the `deprecated` lint.
 - Added exclude option to selene.toml for excluding files from lints
+- Added `testez.yaml` as a supported _Testez_ configuration file.
```
Kampfkarren, line 9:
```suggestion
- Adds support for `.yaml` extensions to be used for standard libraries alongside `.yml`.
```

Your change luckily affects all standard libraries.

selene/src/standard_library.rs
```diff
@@ -72,6 +72,9 @@ fn from_name<V>(
         library = v1_library.into();
     } else {
         let yaml_file = directory.join(format!("{standard_library_name}.yml"));
+        if !yaml_file.exists() {
+            yaml_file = directory.join(format!("{standard_library_name}.yaml"));
+        }
```
Kampfkarren, line 77:
```suggestion
        let mut yaml_file = directory.join(format!("{standard_library_name}.yml"));
        if !yaml_file.exists() {
            yaml_file = directory.join(format!("{standard_library_name}.yaml"));
        }

```

Kampfkarren: The `mut` is necessary, Rust is immutable by default.

Kampfkarren:
Will test this later this week, but this looks good to me. Thanks!

Kampfkarren:
Tested and working, thanks!

Issue #393, opened 2022-07-25
Lint `math.random(1, #x)` in favor of `math.random(#x)`


Kampfkarren:
More concise.

sisco0:
@Kampfkarren, considering the given PR, could you kindly set a review over it? It seems that might be we are missing the fact that `Math.random` could be also called without any parameter for generating a pseudorandom floating point number as mentioned in the [Lua Documentation](http://lua-users.org/wiki/MathLibraryTutorial).

Kampfkarren:
This isn't quite what I meant. `math.random()` is fine, `math.random(5)` is fine, and `math.random(2, 5)` is fine, but `math.random(1, 5)` is the same as `math.random(5)`, and should be linted as such.

Kampfkarren:
I suspect you're going through "good first issue", which is great! This one is gonna be more complicated than the other two--it's a good introduction if you want to get into [writing lints and traversing ASTs](https://kampfkarren.github.io/selene/contributing.html#writing-a-lint), but is not trivial.

sisco0:
By setting a review on the given documentation and current rules files, I got to the conclusion that an `if_chain` should be created with the following elements:

1. If the current instruction is a call to math.random
2. If the number of arguments is 2
3. If the first argument is 1

Thank you for sharing this content with me and helping me to get into development of this application. I would take a more in-depth view and local experimentation before proposing further solutions.

Kampfkarren:
That's correct, with the extra edge case of making sure the user doesn't define `math`. You can do this with the ScopeManager utility, look around for how that's done for other lints.

PR #442, opened 2022-10-03
Add random linter rule for ensuring concise calls

sisco0:
The current PR solves #393 by ensuring that `Math.random` is used with 2 parameters in the linter.
However, it should be taken into account that `Math.random` could also be used without any parameter and returns a floating point number as it could be seen in [Lua Documentation](http://lua-users.org/wiki/MathLibraryTutorial).

A discussion is set in the cited issue about the topic.

### DIFF
--- a/selene-lib/default_std/lua51.yml
+++ b/selene-lib/default_std/lua51.yml
@@ -329,9 +329,8 @@ globals:
     must_use: true
   math.random:
     args:
-      - required: false
-        type: number
-      - required: false
+      - type: number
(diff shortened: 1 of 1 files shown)


Kampfkarren:
To quote my comment on the other one:

> This isn't quite what I meant. `math.random()` is fine, `math.random(5)` is fine, and `math.random(2, 5)` is fine, but `math.random(1, 5)` is the same as `math.random(5)`, and should be linted as such.



Thanks for giving this a try regardless!

PR #443, opened 2022-10-07
chore: use "rule" instead of "lint"

### LINKED ISSUE #48: Replace "rules" with "lints"
Seen both in the source code and the config option. Can deprecate the `rules` config entry but keep it around.

sno2:
Closes #48 

Gotta catch 'em all.


sno2:
Forgot to make the CLI option backwards-compatible, doing that rn

sno2:
Alright, now it's ready for review. Thanks

Kampfkarren:
Nice! Two comments:

1. You should rename RULE_TYPE to LINT_TYPE
2. Some of these changed line endings. My money is that they're being correctly formatted to LF, but I would like if you kept whatever line endings they had before just for sake of review

sno2:
Hmm I've tried both CRLF and LF for `selene-lib/src/lints/unused_variable.rs` and `selene-lib/src/lints/ifs_same_cond.rs` but the diff is not shrinking like others. I am not sure what is going on there.

sno2:
Can you try pasting the code into an online diff checker to make sure I haven't messed with anything else? (https://diffonline.net/ is one)

Kampfkarren:
I would try pulling in the fresh main versions, commit those unchanged, make your replacements, `git diff` beforehand to make sure it's just the rules -> lint changes, and then push.

Also, can you put in the CHANGELOG.md in "Changed" that "rules" in selene.toml should now be "lints", though both are compatible? Eventually we can make this a warning but the infrastructure isn't *quite* there yet.

sno2:
Superseded by #444 . Sorry, couldn't figure out how to get the branches to work together.

PR #449, opened 2022-10-21
Make `assert` message optional

### LINKED ISSUE #448: Don't require error message for `assert`
I use `assert` frequently in my code to assert obvious invariants `#array > 0`, and to assert that something isn't `nil` `assert(value)`, sometimes for typechecking. Writing error messages for these is too verbose, so I simply pass an empty string to get Selene to accept it.

Qualadore:
Closes #448.

### DIFF
--- a/selene-lib/default_std/lua51.yml
+++ b/selene-lib/default_std/lua51.yml
@@ -9,7 +9,7 @@ globals:
   assert:
     args:
       - type: any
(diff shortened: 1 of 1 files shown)


Kampfkarren:
Denied for reasons stated here: https://github.com/Kampfkarren/selene/issues/448#issuecomment-1286642920

PR #452, opened 2022-10-25
Print linting summary as JSON if display-style is json or json2

x3ro:
When requesting "display-style" to be json(2), optimally the entire output of the application should be JSON, so that it can be easily consumed by another application, without having to filter out non-JSON lines. This PR adds the ability to print the summary as JSON instead of printing it as plain text.

PS: Cool project, thanks for building it 🚀

### DIFF
--- a/selene/src/json_output.rs
+++ b/selene/src/json_output.rs
@@ -15,9 +15,9 @@ pub enum JsonOutput {
 
 #[derive(Serialize)]
 pub struct JsonSummary {
-    pub(crate) errors: usize,
-    pub(crate) warnings: usize,
-    pub(crate) parse_errors: usize,
+    errors: usize,
+    warnings: usize,
+    parse_errors: usize,
 }
 
 #[derive(Serialize)]
@@ -104,10 +104,8 @@ pub fn log_total_json(
             errors: lint_errors,
             warnings: lint_warnings,
             parse_errors
-        }))
-            .unwrap()
-    )
-        .unwrap();
+        }))?
+    )?;
 
     Ok(())
 }
--- a/selene/src/json_output.rs
+++ b/selene/src/json_output.rs
@@ -1,12 +1,23 @@
+use std::io;
 use codespan_reporting::diagnostic::{
     Diagnostic as CodespanDiagnostic, Label as CodespanLabel, LabelStyle, Severity,
 };
 use serde::Serialize;
+use termcolor::StandardStream;
+use std::io::Write;
 
 #[derive(Serialize)]
 #[serde(tag = "type")]
 pub enum JsonOutput {
     Diagnostic(JsonDiagnostic),
+    Summary(JsonSummary),
+}
+
+#[derive(Serialize)]
+pub struct JsonSummary {
+    pub(crate) errors: usize,
+    pub(crate) warnings: usize,
+    pub(crate) parse_errors: usize,
 }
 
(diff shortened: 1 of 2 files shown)


selene/src/json_output.rs
```diff
@@ -1,12 +1,23 @@
+use std::io;
 use codespan_reporting::diagnostic::{
     Diagnostic as CodespanDiagnostic, Label as CodespanLabel, LabelStyle, Severity,
 };
 use serde::Serialize;
+use termcolor::StandardStream;
+use std::io::Write;
```
Kampfkarren, line 7:
```suggestion
use std::io::{self, Write};

use codespan_reporting::diagnostic::{
    Diagnostic as CodespanDiagnostic, Label as CodespanLabel, LabelStyle, Severity,
};
use serde::Serialize;
use termcolor::StandardStream;
```

selene/src/json_output.rs
```diff
@@ -1,12 +1,23 @@
...
 #[derive(Serialize)]
 #[serde(tag = "type")]
 pub enum JsonOutput {
     Diagnostic(JsonDiagnostic),
+    Summary(JsonSummary),
+}
+
+#[derive(Serialize)]
+pub struct JsonSummary {
+    pub(crate) errors: usize,
```
Kampfkarren, line 18: No need for `pub(crate)`, just `pub` is acceptable

selene/src/json_output.rs
```diff
@@ -79,3 +90,24 @@ pub fn diagnostic_to_json(
...
+) -> io::Result<()> {
+    writeln!(
+        stdout,
+        "{}",
+        serde_json::to_string(&JsonOutput::Summary(JsonSummary {
+            errors: lint_errors,
+            warnings: lint_warnings,
+            parse_errors
+        }))
+            .unwrap()
```
Kampfkarren, line 108: If we're going to be returning a Result here anyway, these `.unwrap()` should be `?`

selene/src/main.rs
```diff
@@ -23,6 +23,8 @@ use upgrade_std::upgrade_std;
 
 #[cfg(feature = "roblox")]
 use selene_lib::standard_library::StandardLibrary;
+use crate::json_output::log_total_json;
+use crate::opts::DisplayStyle;
```
Kampfkarren, line 27:
```suggestion

use crate::{json_output::log_total_json, opts::DisplayStyle};
```

Kampfkarren (changes requested):
Unfortunately we can only support this on json2, or else existing interpreters of `--display-format json`, which currently only expect outputs to be lints, can break. Can you limit the scope of your changes to just Json2?

Kampfkarren:
BTW let me know if you are doing this for Hacktoberfest. My schedule is somewhat busy so I can't guarantee a merge by Oct 31, but I can definitely give you the accepted label on this PR and the other so that it counts.

x3ro:
That'd be great, re: the hacktoberfest label :) I'll continue to work on the PRs in the coming days!

selene/src/json_output.rs
x3ro: This was a leftover of when I was creating `JsonSummary` in `main`, so I'm going to remove the `pub` completely, in line with the other Json structs.

Kampfkarren:
Thanks!

PR #453, opened 2022-10-25
Add `filename` field to diagnostic messages in JSON output

### LINKED ISSUE #413: JSON output should contain the source name
Source file name is missing from "--display-style Json". The break scenarios where multiple source files are linted.

x3ro:
Fixes #413

### DIFF
--- a/selene/src/json_output.rs
+++ b/selene/src/json_output.rs
@@ -48,7 +48,7 @@ struct Span {
 }
 
 fn label_to_serializable(
-    filename: &String,
+    filename: &str,
     label: &CodespanLabel<codespan::FileId>,
     files: &codespan::Files<&str>,
 ) -> Label {
--- a/selene/src/json_output.rs
+++ b/selene/src/json_output.rs
@@ -32,6 +32,7 @@ pub struct JsonDiagnostic {
 
 #[derive(Serialize)]
 struct Label {
+    filename: String,
     span: Span,
     message: String,
 }
@@ -47,6 +48,7 @@ struct Span {
 }
 
 fn label_to_serializable(
+    filename: &String,
     label: &CodespanLabel<codespan::FileId>,
     files: &codespan::Files<&str>,
 ) -> Label {
@@ -57,6 +59,7 @@ fn label_to_serializable(
         .location(label.file_id, label.range.end as u32)
         .expect("unable to determine end location for label");
     Label {
+        filename: filename.to_string(),
         message: label.message.to_owned(),
         span: Span {
             start: label.range.start,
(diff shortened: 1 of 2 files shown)


x3ro:
In retrospect I actually don't know why I thought this PR would have to depend on #452 🤷🏼‍♂️ It's late now, so I'm going to leave it as is. Let me know if you feel like I should separate the two :)

Kampfkarren:
Separating the two would be helpful--the other PR is going to need a bit more work and I wouldn't want it to delay this one.

x3ro:
Okay, separated the PR from the other one :)

selene/src/json_output.rs
```diff
@@ -62,13 +63,20 @@ pub fn diagnostic_to_json(
     diagnostic: &CodespanDiagnostic<codespan::FileId>,
     files: &codespan::Files<&str>,
 ) -> JsonDiagnostic {
+    let label = diagnostic.labels.first().expect("no labels passed");
+    let filename = files.name(label.file_id)
+        .to_str()
+        .expect("couldn't convert filename to UTF-8")
+        .to_string();
```
Kampfkarren, line 70:
```suggestion
        .to_string_lossy()
        .into_owned();
```

selene/src/json_output.rs
```diff
@@ -11,6 +11,7 @@ pub enum JsonOutput {
 
 #[derive(Serialize)]
 pub struct JsonDiagnostic {
+    filename: String,
```
danielbodorin, line 14: I think filename makes more sense to be part of Span (or Label). This way in the future it can support primary and secondary labels from different files.

Kampfkarren: Ooh that is a good point

Kampfkarren:
Sorry for the delays, I've been really busy :)

Daniel makes a good case about including filename per label, and tests are failing on a format. If you can't make time, I can merge commits into your branch if you want.

x3ro:
Hey! I'll take a look at what this would mean for the implementation sometime during the week, hopefully :) If you get around to it sooner I don't mind, but please let me know here in the ticket, so that we don't work on it in parallel 😅

x3ro:
Finally got around to making this change :) wdyt?

Kampfkarren:
LGTM! Last thing, can you put this in the CHANGELOG.md under the "Added" section of Unreleased? :)

x3ro:
Done 🚀

selene/src/json_output.rs
```diff
@@ -47,6 +48,7 @@ struct Span {
 }
 
 fn label_to_serializable(
+    filename: &String,
```
Kampfkarren, line 51:
```suggestion
    filename: &str,
```

Kampfkarren:
Thanks!

PR #458, opened 2022-11-20
(Fix) Typing of Warn

Barocena:


Kampfkarren:
Needs change log update as well

PR #459, opened 2022-11-22
Add foreachi to 5.1 spec

marcusatiliusregulus:
`table.foreachi` is still available in 5.1, but deprecated:
[deprecation](https://www.lua.org/manual/5.1/manual.html#7.2)
[function signature](https://www.lua.org/manual/5.0/manual.html#5.4)


Kampfkarren:
Looks good, can you update the CHANGELOG as well?

PR #462, opened 2022-11-29
Check config directory for library files

### LINKED ISSUE #339: unable to use standard libraries unless I'm in the same working dir
I've tried using several combinations but I can't figure out how to actually use this linter using a config file and inside neovim using [null-ls.nvim](https://github.com/jose-elias-alvarez/null-ls.nvim/) plugin.

**Attempt 1**: Using `selene --config ~/.config/selene/selene.toml ~/.config/nvim/lua`

The `selene.toml` file contains

```toml
std = "lua51+vim"
```

and another file, `vim.toml` in the same directory contains

```toml
[vim]
any = true
```

When I execute the command mentioned above, I get this output

```
`std = "lua51+vim"`, but some files could not be found:
  `lua51.toml`
  `vim.toml`
ERROR: Could not find all standard library files
```

**Attempt 2**: Using `selene --config ~/.config/selene/selene.toml .config/nvim/lua/`

The `selene.toml` file has the same contents mentioned [here](https://github.com/Kampfkarren/selene/issues/284#issuecomment-892970222)

```toml
# std = "lua51+vim"
[selene]
base = "lua51"

[vim]
any = true
```

This time, selene doesn't recognize the `vim` variable and throws lots of errors like

```
error[undefined_variable]: `vim` is not defined
  ┌─ .config/nvim/lua/autocmd.lua:4:13
  │
4 │ local cmd = vim.cmd
  │             ^^^
```

**Attempt 3**: `cd ~/.config/selene` and `selene ../nvim/lua`

This **succeeds** with the config mentioned in attempt 1 but fails with the configuration in attempt 2. Does that mean `selene.toml` should always be in the same working dir wherever selene command is run from? If yes, what purpose does `--config` serve?

**Attempt 4**: `selene.toml` and `vim.toml` are in `~/.config/nvim/lua/cfg`, the dir where `null-ls` is configured, `~/.config/nvim/lua/`, and `~/.config/nvim/` - one after the other

`null-ls` was configured not to use `--config` in this case and I see errors related to `vim` not being present using configs in both attempt 1 and 2

The only case where `null-ls` and `selene` don't complain about `vim` and **succeed** is when I execute `selene` or open neovim while being in the same dir wherever `selene.toml` is with the config mentioned in attempt 1.

I don't think it's feasible to create `selene.toml` and `vim.toml` in each and every folder on my operating system? How do I actually configure this linter to use the file in `~/.config/selene/selene.toml` with the contents in attempt 1?

deathlyfrantic:
I'm sure this code could be improved - my Rust is a bit rusty (ha). Happy to address any issues.

Should fix #339.

### DIFF
--- a/selene/src/main.rs
+++ b/selene/src/main.rs
@@ -435,79 +435,87 @@ fn start(mut matches: opts::Options) {
         None => {}
     }
 
-    let config: CheckerConfig<toml::value::Value> = match matches.config {
-        Some(config_file) => {
-            let config_contents = match fs::read_to_string(config_file) {
-                Ok(contents) => contents,
-                Err(error) => {
-                    error!("Couldn't read config file: {}", error);
-                    std::process::exit(1);
-                }
-            };
+    let (config, config_directory): (CheckerConfig<toml::value::Value>, Option<PathBuf>) =
+        match matches.config {
+            Some(config_file) => {
+                let config_contents = match fs::read_to_string(&config_file) {
+                    Ok(contents) => contents,
+                    Err(error) => {
+                        error!("Couldn't read config file: {}", error);
+                        std::process::exit(1);
+                    }
+                };
 
-            match toml::from_str(&config_contents) {
-                Ok(config) => config,
-                Err(error) => {
-                    error!("Config file not in correct format: {}", error);
-                    std::process::exit(1);
+                match toml::from_str(&config_contents) {
+                    Ok(config) => (
+                        config,
+                        Path::new(&config_file).parent().map(Path::to_path_buf),
+                    ),
+                    Err(error) => {
+                        error!("Config file not in correct format: {}", error);
+                        std::process::exit(1);
+                    }
                 }
             }
-        }
 
-        None => match fs::read_to_string("selene.toml") {
-            Ok(config_contents) => match toml::from_str(&config_contents) {
-                Ok(config) => config,
-                Err(error) => {
-                    error!("Config file not in correct format: {}", error);
-                    std::process::exit(1);
-                }
+            None => match fs::read_to_string("selene.toml") {
+                Ok(config_contents) => match toml::from_str(&config_contents) {
+                    Ok(config) => (config, None),
+                    Err(error) => {
+                        error!("Config file not in correct format: {}", error);
+                        std::process::exit(1);
+                    }
+                },
+
+                Err(_) => (CheckerConfig::default(), None),
             },
-
-            Err(_) => CheckerConfig::default(),
-        },
-    };
+        };
 
     let current_dir = std::env::current_dir().unwrap();
 
-    let standard_library =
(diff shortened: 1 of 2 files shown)


deathlyfrantic:
I ran rustfmt on the two files I changed before committing but somehow still had errors in the `cargo fmt --check` run. So I did a `cargo fmt` and that changed every line in `main.rs`, I'm guessing because of line endings? What a mess.

selene/src/main.rs
```diff
@@ -435,79 +435,87 @@ fn start(mut matches: opts::Options) {
...
 
-            match toml::from_str(&config_contents) {
-                Ok(config) => config,
-                Err(error) => {
-                    error!("Config file not in correct format: {}", error);
-                    std::process::exit(1);
+                match toml::from_str(&config_contents) {
+                    Ok(config) => (
+                        config,
+                        Path::new(&config_file).parent().map(|p| p.to_path_buf()),
```
Kampfkarren, line 452:
```suggestion
                        Path::new(&config_file).parent().map(Path::to_path_buf),

```

selene/src/standard_library.rs
```diff
@@ -59,39 +63,55 @@ fn from_name<V>(
...
     directory: &Path,
+    config_dir: &Option<PathBuf>,
 ) -> color_eyre::Result<Option<StandardLibrary>> {
-    let mut library: StandardLibrary;
+    let mut library: Option<StandardLibrary> = None;
 
-    let toml_file = directory.join(format!("{standard_library_name}.toml"));
-    if toml_file.exists() {
-        let content = fs::read_to_string(&toml_file)?;
+    let mut dirs = vec![directory];
```
Kampfkarren, line 70: Full name please, `directories`

selene/src/standard_library.rs
```diff
@@ -59,39 +63,55 @@ fn from_name<V>(
...
+    if let Some(dir) = config_dir {
+        dirs.push(dir);
+    };
 
-        let v1_library: v1::StandardLibrary = toml::from_str(&content)
-            .with_context(|| format!("failed to read {}", toml_file.display()))?;
+    for dir in dirs {
```
Kampfkarren, line 75: `directory`

Kampfkarren (changes requested):
Small nits

selene/src/standard_library.rs
deathlyfrantic: I changed this and also all the instances of `config_dir` to `config_directory`.

deathlyfrantic: Although maybe I should've left `config_dir` since you're already using `current_dir`. 🤔

deathlyfrantic:
Not sure what's going on with [this test failure](https://github.com/Kampfkarren/selene/actions/runs/3639114087/jobs/6142113590). I don't think I changed anything that would've caused it. 🤔

Kampfkarren:
Yeah that one is my fault :(

PR #481, opened 2023-01-18
fix error label range, bump deps

attila-lin:
and fix clippy

### DIFF
--- a/selene-lib/tests/lints/mismatched_arg_count/reassigned_variables_2.stderr
+++ b/selene-lib/tests/lints/mismatched_arg_count/reassigned_variables_2.stderr
@@ -1,5 +1,5 @@
 error[mismatched_arg_count]: this function takes 3 arguments but 4 arguments were supplied
-   ┌─ reassigned_variables_2.lua:1:16
+   ┌─ reassigned_variables_2.lua:11:1
    │
  1 │ local function foo(a, b, c)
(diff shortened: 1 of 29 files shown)


selene-lib/tests/lints/mismatched_arg_count/reassigned_variables_2.stderr
```diff
@@ -11,7 +11,7 @@ error[mismatched_arg_count]: this function takes 3 arguments but 4 arguments wer
    │ ^^^^^^^^^^^^^^^ expected 3 arguments
 
 error[mismatched_arg_count]: this function takes 3 arguments but 4 arguments were supplied
-   ┌─ reassigned_variables_2.lua:1:16
+   ┌─ reassigned_variables_2.lua:13:1
```
Kampfkarren, line 14: I don't understand, these are changing to `13:1`, but the line number in the stderr still says `1`.

attila-lin: image


Can't reproduce.
Working my env right. Maybe confuse with the `warning` part?

Kampfkarren: Maybe we're just passing the wrong things into codespan then :-(

PR #486, opened 2023-02-17
Add CFrame.fromEulerAngles API

DarraghGriffin:
Add the CFrame.fromEulerAngles API which takes Enum.RotationOrder as an optional fourth argument.


Kampfkarren:
:+1: Add a changelog entry and we're all set

PR #492, opened 2023-02-20
Show config errors in the VS Code extension

### LINKED ISSUE #392: VS Code extension does not tell you when your standard library/config is wrong

Kampfkarren:
Closes #392


selene/src/validate_config.rs
```diff
@@ -0,0 +1,123 @@
...
+impl From<serde_yaml::Location> for ErrorRange {
+    fn from(location: serde_yaml::Location) -> Self {
+        Self {
+            start: location.index(),
+            end: location.index(),
+        }
+    }
+}
+
+// TODO: Test
```
matthargett, line 48: Why not add the test in the same PR?

matthargett (changes requested):
Lots of TS file changes, but no tests covering the changes. You could do a pre-release without tests to elicit focused feedback, but without tests this seems like a lot to push to users.

selene/src/validate_config.rs
Kampfkarren: All the TODOs will be solved before I undraft the PR, but I knew I'd forget if I didn't write it down

Kampfkarren:
I think the VS Code extensions have been ridiculously hard to even make simple tests for, I'll see what I can do

Kampfkarren:
Going to ship a pre-release version of the VSC extension when this is ready

Kampfkarren:
Forgot changelog, will commit that later

Issue #433, opened 2022-09-26
Properly add Instance.new's second parameter to the standard library and mark it deprecated


Kampfkarren:
Right now, `Instance.new("Class", parent)` creates an error, as the second parameter for Instance.new is not in the standard library. This is intentional, as per https://devforum.roblox.com/t/psa-dont-use-instancenew-with-parent-argument/30296 . While there are cases where it is okay, every case where it's okay can be replaced with just using `.Parent = parent` afterwards. 

However, doing it this way causes several problems.

One is that it isn't correct, and causes selene to overblow the issue when linting. `Instance.new("Class", parent)`, which works but is not recommended, is reported as an error just like `math.thisdoesnotexist()`. 

The second is by bundling it with the standard library, it causes a lot of people to disable *all* of incorrect_standard_library_use, which you should never ever ever do.

The third is that it makes it unclear *why* you should not be doing this. This is one of the most common trip ups Roblox developers get, and it has been incorrectly reported as a bug several times. There should be a clear message explaining what the problem is, and link to the developer forums thread.

Currently impeding adoption within Roblox, where this has unfortunately come up a lot.

Kampfkarren:
Got confirmation that the second parameter is deprecated. This will probably need to come with the ability to block specific deprecations without disabling the whole lint.

ambergamefam:
Could the output message on this be improved to say something like "The second parameter for 'Instance.new' is deprecated. Please assign the instance's Parent in a separate statement."?

It's confusing since it's not actually a type error in Luau / Roblox Studio, but emits from selene as though it were one.

Kampfkarren:
`deprecated` is warn by default (as opposed to standard library's error) and allows custom error messages, so yes.

JohnMaxwellDistinti:
Considered making a PR for this after noticing the warning. Thank you for your explanation!

PR #496, opened 2023-03-18
Allow second argument in `Instance.new`

ccuser44:
It is not deprecated and it's quite odd that it's disabled.
Also it's not bad usage. It's only bad usage if properties are set after. If properties are not set after setting the Parent then there is no issue.


Kampfkarren:
It is deprecated, I have already confirmed this with some engineers. https://github.com/Kampfkarren/selene/issues/433 is the only proper way to do this.

PR #508, opened 2023-04-16
Lint `Name` property in Roact

### LINKED ISSUE #245: Lint `Name` property in Roact

chriscerie:
Closes  #245

### DIFF
--- a/selene-lib/src/lints/roblox_incorrect_roact_usage.rs
+++ b/selene-lib/src/lints/roblox_incorrect_roact_usage.rs
@@ -17,18 +17,30 @@ use if_chain::if_chain;
 
 pub struct IncorrectRoactUsageLint;
 
+// Assumes string includes quotes at start and end
+fn is_lua_valid_table_key_identifier(string: &String) -> bool {
+    // Valid identifier cannot start with numbers
+    let first_char = string.chars().nth(1).unwrap();
+    if !first_char.is_alphabetic() && first_char != '_' {
+        return false;
+    }
+
+    string
+        .chars()
+        .skip(1)
+        .take(string.len() - 2)
+        .all(|c| c.is_alphanumeric() || c == '_')
+}
+
 fn get_lua_table_key_format(expression: &ast::Expression) -> String {
     match expression {
         ast::Expression::Value { value, .. } => match &**value {
             ast::Value::String(token) => {
                 let string = token.to_string();
-                if string
-                    .char_indices()
-                    .any(|(i, c)| i != 0 && i != string.len() - 1 && !c.is_alphabetic())
-                {
-                    format!("[{}]", string)
-                } else {
+                if is_lua_valid_table_key_identifier(&string) {
                     string[1..string.len() - 1].to_string()
+                } else {
+                    format!("[{}]", string)
                 }
             }
             _ => format!("[{}]", expression.to_string()),
--- a/selene-lib/src/lints/roblox_incorrect_roact_usage.rs
+++ b/selene-lib/src/lints/roblox_incorrect_roact_usage.rs
@@ -142,7 +142,7 @@ struct InvalidEvent {
 struct InvalidProperty {
     class_name: String,
     property_name: String,
-    property_value: Box<ast::Expression>,
+    property_value: ast::Expression,
     create_element_expression: String,
     range: (usize, usize),
 }
@@ -245,7 +245,7 @@ impl<'a> Visitor for IncorrectRoactUsageVisitor<'a> {
                         self.invalid_properties.push(InvalidProperty {
                             class_name: name.clone(),
                             property_name,
-                            property_value: Box::new(value.clone()),
+                            property_value: value.clone(),
                             create_element_expression: create_element_expression.clone().to_owned(),
                             range: range(key),
                         });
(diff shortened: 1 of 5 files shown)


selene-lib/src/lints/roblox_incorrect_roact_usage.rs
```diff
@@ -65,14 +65,28 @@ impl Lint for IncorrectRoactUsageLint {
...
-                    invalid_property.property_name, invalid_property.class_name
-                ),
-                Label::new(invalid_property.range),
-            ));
+            match invalid_property.property_name.as_str() {
+                "Name" => {
+                    diagnostics.push(Diagnostic::new(
+                        "roblox_incorrect_roact_usage",
+                        format!(
+                            "`{}` is assigned through the element's key for host components",
```
Kampfkarren, line 73: "Host component" is a confusing internal term. This instead ought to be just "Roblox instances"

selene-lib/tests/lints/roblox_incorrect_roact_usage/roblox_incorrect_roact_usage.stderr
```diff
@@ -28,3 +28,9 @@ error[roblox_incorrect_roact_usage]: `ThisDoesntExist` is not a valid class
 27 │ e("ThisDoesntExist")
    │   ^^^^^^^^^^^^^^^^^
 
+error[roblox_incorrect_roact_usage]: `Name` is assigned through the element's key for host components
+   ┌─ roblox_incorrect_roact_usage.lua:30:5
+   │
+30 │     Name = "Should not be passed",
```
Kampfkarren, line 34: This should come with a `try: ["Should not be passed"] = e(` if it's in a createElement call of its own.

Kampfkarren (changes requested):
CHANGELOG needed

chriscerie:
The help note now uses the `createElement` expression that you're using.

```lua
e("Frame", {
    Name = "Should not be passed", -- try: ["Should not be passed"] = e(...)
})

Roact.createElement("Frame", {
    Name = a, -- try: [a] = Roact.createElement(...)
})
```

selene-lib/src/lints/roblox_incorrect_roact_usage.rs
```diff
@@ -136,6 +142,8 @@ struct InvalidEvent {
 struct InvalidProperty {
     class_name: String,
     property_name: String,
+    property_value: Box<ast::Expression>,
```
Kampfkarren: Why does this need to be boxed? If it's just because we're taking it from something that's boxed, deref first.

selene-lib/tests/lints/roblox_incorrect_roact_usage/roblox_incorrect_roact_usage.stderr
```diff
@@ -28,9 +28,19 @@ error[roblox_incorrect_roact_usage]: `ThisDoesntExist` is not a valid class
...
-error[roblox_incorrect_roact_usage]: `Name` is assigned through the element's key for host components
-   ┌─ roblox_incorrect_roact_usage.lua:30:5
+error[roblox_incorrect_roact_usage]: `Name` is assigned through the element's key for Roblox instances
+   ┌─ roblox_incorrect_roact_usage.lua:38:5
    │
-30 │     Name = "Should not be passed",
+38 │     Name = "Should not be passed",
    │     ^^^^
+   │
+   = try: ["Should not be passed"] = e(...)
```
Kampfkarren, line 37: Can we add a case too for a name that can be passed in as just `Whatever = e`? We shouldn't be suggesting `[]` unless we need to.

Kampfkarren (changes requested):
Better!

chriscerie:
New behavior:

```lua
e("Frame", {
    Name = "Should not be passed", -- try: ["Should not be passed"] = e(...)
})

e("Frame", {
    Name = "ShouldNotBePassed", -- try: ShouldNotBePassed = e(...)
})
```

chriscerie:
Also included the case of periods.

```lua
e("Frame", {
    Name = "Should.not.be.passed", -- try: ["Should.not.be.passed"] = e(...)
})
```

chriscerie:
I realized that the original omit logic would false positive in some cases where it starts with a number. Now:

```lua
-- Includes []
"Should not be passed"
"Should.not.be.passed"
"0Should0not0be0passed"

-- Omits []
"Should0not0be0passed"
"_Should_not_be_passed"
"ShouldNotBePassed"
```

selene-lib/src/lints/roblox_incorrect_roact_usage.rs
```diff
@@ -17,6 +17,23 @@ use if_chain::if_chain;
 
 pub struct IncorrectRoactUsageLint;
 
+fn get_lua_table_key_format(expression: &ast::Expression) -> String {
+    match expression {
+        ast::Expression::Value { value, .. } => match &**value {
+            ast::Value::String(token) => {
+                let string = token.to_string();
+                if string.contains(" ") {
```
Kampfkarren: Not just spaces. Needs to be a valid identifier.

Test cases:

"A-B-C", should become []
"ABC", should become ABC
"ABC123", should become ABC123
"123ABC", should become []

chriscerie: Sorry, realized that after requesting review, fixed in https://github.com/Kampfkarren/selene/pull/508/commits/58fa9ee22d532d0f37e659b48da2c2bbb5562de7.

Kampfkarren:
A few real clippy lints coming through

Kampfkarren:
Great, thanks!

PR #510, opened 2023-04-19
Add `empty_loop` lint

### LINKED ISSUE #509: Warn against looping over a table that is never used
```lua
local things = {}

for _, thing in things do
```

...should lint.

### LINKED ISSUE #511: Warn against empty loop
Empty loops like `while true do end` should lint due to the reasons written in [empty_if](https://kampfkarren.github.io/selene/lints/empty_if.html).

chriscerie:
Closes #511.

This also lints for loops not involving tables, like
```lua
while true do end
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - Added `table.move`to Lua 5.3 standard library
 - Added `bit32.*` functions to Lua 5.2 standard library
 - Added `table.pack`, `rawlen` and `package.config` to Lua 5.2 standard library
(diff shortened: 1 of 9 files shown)


Kampfkarren:
This is fine, but it doesn't close #509. The bug I ran into there was I had a table, never did anything with it that would have led it to insert anything, then tried to loop over it, which would never have worked and we could statically analyze it.

chriscerie:
> This is fine, but it doesn't close #509. The bug I ran into there was I had a table, never did anything with it that would have led it to insert anything, then tried to loop over it, which would never have worked and we could statically analyze it.

Ah I misunderstood the issue. I created a new issue to track this.

Kampfkarren:
Thanks!

PR #518, opened 2023-04-29
Fix undefined_variable false negative when defining multiple methods in undefined tables

### LINKED ISSUE #67: undefined_variable doesn't show up for function declarations
```lua
function Foo:bar()
end
```

...does not lint if Foo does not exist.

### LINKED ISSUE #516: Adding a method twice for an undefined table no longer errors
This has been broken for a while, but basically if you have a table defined with the same method twice it, won't show an error that the table `DOES_NOT_EXIST` does not exist. 

```lua
--[=[
	A BaseObject basically just adds the :Destroy() interface, and a _maid, along with an optional object it references.
	@class BaseObject
]=]

local require = require(script.Parent.loader).load(script)

local Maid = require("Maid")

local BaseObject = {}
BaseObject.ClassName = "BaseObject"
BaseObject.__index = BaseObject

--[=[
	Constructs a new BaseObject

	@param obj? Instance
	@return BaseObject
]=]
function BaseObject.new(obj)
	local self = setmetatable({}, BaseObject)

	self._maid = Maid.new()
	self._obj = obj

	return self
end

function DOES_NOT_EXIST:DoThing()

end

function DOES_NOT_EXIST:DoOtherThing()

end

--[=[
	Cleans up the BaseObject and sets the metatable to nil
]=]
function BaseObject:Destroy()
	self._maid:DoCleaning()
	setmetatable(self, nil)
end

return BaseObject
```

chriscerie:
Fixes #67 and fixes #516. It looks like `is_longer_expression` only checked for `function a.b()`, but didn't account for method colons like `function a:b()`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 ### Fixed
 - `string.pack` and `string.unpack` now have proper function signatures in the Lua 5.3 standard library.
 - Moved `math.log` second argument addition from Lua 5.3 std lib to 5.2 std lib
+- `undefined_variable` now correctly errors when defining multiple methods in undefined tables
 
 ## [0.25.0](https://github.com/Kampfkarren/selene/releases/tag/0.25.0) - 2023-03-12
 ### Added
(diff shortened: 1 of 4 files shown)


Kampfkarren:
Thanks!

PR #519, opened 2023-05-02
Add ignore_pattern config to global_usage

### LINKED ISSUE #181: feature request: allow ignore_pattern for global_usage
If would be nice if I could have global_usage not warn on safe, read-only globals like this:
```
global_usage = { ignore_pattern = "__DEV__|__EXPERIMENTAL__" }
```

Which would then not warn on code like this:
```
if _G.__DEV__ then 
  -- ...
end
```

But would still warn on code like this:
```
if _G.__MUTABLE_STATE_FTW then
  -- ...
end
```

chriscerie:
Closes #181.


matthargett (approved):

selene-lib/src/lints/global_usage.rs
```diff
@@ -1,23 +1,43 @@
...
 
 fn is_global(name: &str, roblox: bool) -> bool {
     (roblox && name == "shared") || name == "_G"
 }
 
-pub struct GlobalLint;
+#[derive(Clone, Deserialize)]
+#[serde(default)]
+pub struct GlobalConfig {
+    ignore_pattern: String,
```
Kampfkarren, line 15: Instead of doing `.^`, we should instead make this `Option<String>` and save the performance IMO.

Kampfkarren (approved):

docs/src/lints/global_usage.md
```diff
@@ -6,7 +6,7 @@ Prohibits use of `_G`.
 `_G` is global mutable state, which is heavily regarded as harmful. You should instead refactor your code to be more modular in nature.
 
 ## Configuration
-`ignore_pattern` (default: `".^"`) - A [regular expression](https://en.wikipedia.org/wiki/Regular_expression) for variables that are allowed to be global variables. The default disallows all global variables regardless of their name.
+`ignore_pattern` (default: `None`) - A [regular expression](https://en.wikipedia.org/wiki/Regular_expression) for variables that are allowed to be global variables. The default disallows all global variables regardless of their name.
```
Kampfkarren, line 9:
```suggestion
`ignore_pattern` - A [regular expression](https://en.wikipedia.org/wiki/Regular_expression) for variables that are allowed to be global variables. The default disallows all global variables regardless of their name.
```

selene-lib/src/lints/global_usage.rs
```diff
@@ -36,7 +28,9 @@ impl Lint for GlobalLint {
 
     fn new(config: Self::Config) -> Result<Self, Self::Error> {
         Ok(GlobalLint {
-            ignore_pattern: Regex::new(&config.ignore_pattern)?,
+            ignore_pattern: config
+                .ignore_pattern
+                .and_then(|ignore_pattern| Regex::new(&ignore_pattern).ok()),
```
Kampfkarren, line 33: We're ignoring the error!

PR #521, opened 2023-05-18
Add react lint

### LINKED ISSUE #484: Support React for incorrect Roact usage lint
- Track the word "React"
- When "React" is the name of the variable, include "ref" as a valid property

chriscerie:
Closes #484.

The alternative implementation is to bundle React logic in the existing `roblox_incorrect_roact_usage` lint to reduce duplicated code. I decided to make a new lint for React so separating roact and react-lua logic is more sane as more rules specific to react-lua is added. But it's an easy fix if we want to merge the React rules into the existing roact lint.

Note the layout in the doc. Since this includes 3 rules, I found organizing this by rules is more readable, but it's also an easy change if we prefer consistency with other existing docs.


chriscerie:
Ended up bundling react rules into the existing `roblox_incorrect_roact_usage` following a brief discussion with @Kampfkarren.

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - Added new [`roblox_suspicious_udim2_new` lint](https://kampfkarren.github.io/selene/lints/roblox_suspicious_udim2_new.html), which will warn when you pass in too few number of arguments to `UDim2.new`.
 - `roblox_incorrect_roact_usage` now lints for illegal `Name` property
 - Added `ignore_pattern` config to `global_usage`, which will ignore any global variables with names that match the pattern
+- `roblox_incorrect_roact_usage` now supports react-lua
```
Kampfkarren, line 13: 1. Everywhere you refer to this as "react-lua", call it Roact17. It's the name I think everyone should probably start using
2. Be more specific about what this entails--so that `key`, `children`, and `ref` are no longer linted against, and that `React` is accepted as a variable name

Kampfkarren (approved):

PR #530, opened 2023-05-31
Add force-exclude command-line flag

captalbator:
Adds a `force-exclude` flag, which forces selene to ignore files that match the globs in `exclude` even when ran on a single file.

This is mostly intended for the vscode extension, as it is a common complaint that the extension does not respect excludes.


Kampfkarren:
I think this should just be the native behavior, I don't think other linters work this way.

captalbator:
Alright. I'll change the behavior and add a --no-exclude flag instead.

PR #533, opened 2023-06-26
Add Roact dangling connection lint

### LINKED ISSUE #383: Calling :Connect as a statement in a Roact component should lint
Roact components can be mounted and unmounted regularly. Connecting to an event as a statement means you can't disconnect it, creating memory leaks or use after unmount. 

Would necessarily be a bit finicky, since it can only check you're calling *something* with :Connect, and there are likely APIs with their own unrelated Connect method, or weird maids-within-callbacks type deal, but these are likely extremely uncommon and would be worth the author just filtering them out.

chriscerie:
Closes #383. This only errors when the file has `Roact` or `React` defined and when the connection is made within useEffect. In the future if we can reliably detect components vs regular functions we can also warn about dangling connections during render.

```lua
a:Connect()

local function a()
    b:Connect()

    useEffect(function()
        local b = c:Connect()
        d(b:Connect())
        e(function() end, f:Connect())

        c:Connect() -- Errors
    end, {})
end
```


selene-lib/src/lints/roblox_roact_dangling_connection.rs
```diff
@@ -0,0 +1,234 @@
...
+            function_contexts: Vec::new(),
+        };
+
+        visitor.visit_ast(ast);
+
+        let mut diagnostics = Vec::new();
+
+        for invalid_event in visitor.dangling_connections {
+            match invalid_event.function_context {
+                ConnectionContext::UseEffect => {
```
Kampfkarren, line 68: `if let` instead

Kampfkarren (approved):
Looks mostly good. Out of curiosity, and it doesn't have to, does this magically work with ReactHooks' useEffect? The one being passed through?

chriscerie:
> Looks mostly good. Out of curiosity, and it doesn't have to, does this magically work with ReactHooks' useEffect? The one being passed through?

Yes, it will show the useEffect-specific error with `useEffect()` and `anything.useEffect()`

chriscerie:
To display the useEffect specific error it'll now look for the react/roact/hooks prefix per @matthargett's suggestion for a similar lint

PR #543, opened 2023-08-12
Attempt to find `selene.toml` in parent directories

### LINKED ISSUE #526: Subfolder selene.toml
## Explanation

I would like selen to handle selene.toml's in subfolders.

## Example

```lua
📦project
 ┣ 📂spec
 ┃ ┣ 📜test_spec.lua -- should use /spec/selene.toml
 ┃ ┣ 📄busted.selene.yaml
 ┃ ┗ ⚙️selene.toml
 ┣ 📂src
 ┃ ┣ 📄cc-tweaked.selene.yaml
 ┃ ┣ 📜test.lua -- should use /src/selene.toml
 ┃ ┗ ⚙️selene.toml
 ┣ 📂some_folder
 ┃ ┗ 📜test.lua -- should use /selene.toml
 ┣ 📜test.lua -- should use /selene.toml
 ┗ ⚙️selene.toml
```

ProspectPyxis:
With this change, if the `--config` flag is not specified, selene will search the current directory and all parent directories for a `selene.toml` file, rather than only looking in the current directory.

This does *not* affect standard library file resolution - it is assumed that the standard library file is in the same directory as the config file.

Closes #526.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,9 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - Added `ignore_pattern` config to `global_usage`, which will ignore any global variables with names that match the pattern
 - `roblox_incorrect_roact_usage` now checks for incorrect Roact17's `createElement` usage on variables named `React`. For Roact17 only, `key`, `children`, and `ref` are valid properties to Roblox instances.
 
+### Changed
+- If `--config` is not defined, `selene.toml` will now be searched for in the current directory and all parent directories.
+
 ### Fixed
 - `string.pack` and `string.unpack` now have proper function signatures in the Lua 5.3 standard library.
 - Moved `math.log` second argument addition from Lua 5.3 std lib to 5.2 std lib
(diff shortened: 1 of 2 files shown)


chriscerie:
This behavior should be added to docs. This is good, but this serves a different use case than #526.

ProspectPyxis:
> I’m concerned about the case if the user accidentally adds a `selene.toml` in some ancestor directory, it may be difficult to debug why selene is not behaving in the way they expect. Should it output a message that it’s getting the config file in an ancestor directory?

I don't think this will be a problem - many other linters/formatters work the same way when locating their config files. If we're concerned for backwards compatibility, though, we could make this a `--parent-config` command line flag instead or something along those lines.

chriscerie:
> If we're concerned for backwards compatibility, though, we could make this a `--parent-config` (name not final) command line flag.

My intuition says such cases where a config file exists somewhere higher but the user relies on default config values is much more likely locally (which an output would make it obvious why it's behaving differently) than in a controlled ci environment.

But at the very least document this behavior and link the pr to a relevant issue (not #526).

cxwx:
still not merged？
I use ale for linting，but the neovim always change current dir when open directory.
So I really need the feature.

PR #545, opened 2023-08-15
Show missing required param message when in standard library format

### LINKED ISSUE #544: Specify message passed to standard library format for standard_library missing required parameter
With standard library format:
```yml
assert:
  args:
    - type: any
    - required: A failed assertion without a message is unhelpful to users.
      type: string
```

Currently if user specifies only one arg:
```
error[incorrect_standard_library_use]: standard library function `assert` requires 2 parameters, 1 passed
  ┌─ assert.lua:6:1
  │
6 │ assert(true)
  │ ^^^^^^^^^^^^
  │
  ```
This has been the source of confusion for many as the 2nd `assert` parameter is technically optional, but selene chooses to enforce it as required since the default assert error is unhelpful. The current selene error doesn't specify that, so the message in the standard library format should be added as a note.

Another example of this is with `error` to ensure solution is generic:
```yml
error:
  args:
    - required: Erroring without a message is unhelpful to users.
      type: string
    - required: false
      type: number
```

chriscerie:
Closes #544.

If the function has multiple required parameters each with their own message, this will only show the first missing param if it has a message.

The alternative to this is to show every missing parameter's message with an indicator to which arg they associate with, but this may be confusing in cases like `assert()` with no args. The error message would show the standard library format message of `A failed assertion without a message is unhelpful to users`, but the bigger issue is they aren't asserting anything at all.


Kampfkarren:
Weird that this wasn't tested, thanks!

PR #549, opened 2023-08-27
Remove version constraints on container images

f1rstlady:
This fixes errors during the build of this image, occuring due to non-existent version tags.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,20 +1,20 @@
-FROM rust:1.64-bullseye AS selene-builder
+FROM rust:bullseye AS selene-builder
 RUN apt-get update && \
     apt-get upgrade -y && \
     apt-get install g++ && \
     cargo install --branch main --git https://github.com/Kampfkarren/selene selene
 
-FROM rust:1.64-bullseye AS selene-light-builder
+FROM rust:bullseye AS selene-light-builder
 RUN apt-get update && \
     apt-get upgrade -y && \
     apt-get install g++ && \
     cargo install --no-default-features --branch main --git https://github.com/Kampfkarren/selene selene
 
-FROM rust:1.64-alpine3.14 AS selene-musl-builder
+FROM rust:alpine AS selene-musl-builder
 RUN apk add g++ && \
     cargo install --branch main --git https://github.com/Kampfkarren/selene selene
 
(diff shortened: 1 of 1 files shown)


Jarmos-san (changes requested):
Can you make use [build-time variables](https://docs.docker.com/engine/reference/commandline/build/#build-arg) instead? That would allow for more flexibility to build the Docker Image. See #550 for some ideas I had in mind.

f1rstlady:
> Can you make use [build-time variables](https://docs.docker.com/engine/reference/commandline/build/#build-arg) instead? That would allow for a more flexible to build the Docker Image. See #550 for some ideas I had in mind.

Do we need this flexibility? I'm asking because I prefer to add the minimal needed abstraction only. In addition, I suggest to distinguish between _fixing_ and _enhancing_ the image.

Other ideas for enhancements:

1. Do we need the debian-based builds? Afaik, docker is used only by the pre-commit hook, which in turn uses the alpine-based image only.
2. @Kampfkarren [As the build pipeline shows](https://github.com/Kampfkarren/selene/blob/b9c27b0d4de9c7cb84b111ed58f825143bb104a0/.github/workflows/build.yml#L198), the built images are pushed to the Docker hub. But under which account? Shall we document the location of the builds?

Jarmos-san:
> Do we need this flexibility? I'm asking because I prefer to add the minimal needed abstraction only. In addition, I suggest to distinguish between fixing and enhancing the image.

Don't fix what's not broken is definitely the easier way around the issue with the Docker build issue right now tbh. The flexible enhancements I suggested on the other hand is to future-proof a potential scenario wherein someone had to test the build against a different Rust Docker Image for whatever reason.

For example, what if someone had to quickly check if a certain feature works on a particular Image or perhaps a specific version of Rust? Passing the build-time variables to `docker` would allow us to do that.

The other intention I had in mind was, since the current `Dockerfile` is worked on, why not do it the proper way right now instead of waiting for someone else to do it in the future? It'll allow Kamp to be relieved off of some maintenance burden as well.

Jarmos-san (approved):

Kampfkarren:
@f1rstlady My assumption for the account is going to be `boyned`. This looks fine

PR #565, opened 2023-10-30
Add pre-commit hook

### LINKED ISSUE #539: Support for pre-commit hooks?
There is this tool called [pre-commit](https://pre-commit.com) which supports creating and managing multi-language pre-commit hooks easily. I believe you only need to add a `.pre-commit-hooks.yaml` file to the repository and the `pre-commit` tool will take care of compiling and invoking the binary when necessary! I use Lua (and Selene) both at work and for personal projects besides other languages (and tools) hence having one single tool to manage our pre-commit hooks is necessary.

There is more information in the "[Creating New Hooks](https://pre-commit.com/#new-hooks)" section of the docs. If you are willing to support pre-commit hooks for Selene, I think I can write a config for the hook and share a PR. So let me know your thoughts?

amitds1997:
Fixes #539

Rust-based pre-commit does not work because pre-commit hardcodes the `--path` to `.` which leads to build failure and errors with `found a virtual manifest at <path>/Cargo.toml` instead of a package manifest. Pre-commit maintainer disagrees with making this configurable so it is not possible to add a `rust` based hook: https://github.com/pre-commit/pre-commit/issues/2931

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -1,0 +1,21 @@
+---
+- id: selene-system
+  name: selene (system)
+  description: An opinionated Lua code linter
+  entry: selene
+  language: system
+  types: [lua]
+
+- id: selene-docker
+  name: selene (docker)
+  description: An opinionated Lua code linter
+  entry: /selene
+  language: docker
+  types: [lua]
+
+- id: selene-github
+  name: selene (GitHub)
+  description: An opinionated Lua code linter
+  entry: selene
+  language: python
+  types: [lua]
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - Added `Path2DControlPoint.new` to the Roblox standard library
 - [Adds `lua_versions` to standard library definitions](https://kampfkarren.github.io/selene/usage/std.html#lua_versions). Specifying this will only allow the syntax used by those languages. The default standard libraries now specify these, meaning that invalid syntax for that language will no longer be supported.
 - Added missing third parameter to `PathWaypoint.new` in the Roblox standard library
+- Added support for selene as a `pre-commit` hook
 
 ### Changed
 - Upgrades to [full-moon 1.0.0](https://github.com/Kampfkarren/full-moon/blob/main/CHANGELOG.md#100---2024-10-08), which should provide faster parse speeds, support for multiple parsing errors at the same time, and support for some new Luau syntax.
--- a/docs/src/SUMMARY.md
+++ b/docs/src/SUMMARY.md
@@ -10,6 +10,7 @@
   - [Configuration](./usage/configuration.md)
   - [Filtering](./usage/filtering.md)
   - [Standard Library Format](./usage/std.md)
+  - [Git hooks (pre-commit)](./usage/git-hooks-pre-commit.md)
 - [Roblox Guide](./roblox.md)
 - [Contributing](./contributing.md)
 - [Lints](./lints/index.md)
--- a/docs/src/usage/git-hooks-pre-commit.md
+++ b/docs/src/usage/git-hooks-pre-commit.md
@@ -1,0 +1,35 @@
+# Pre-commit
+
+`pre-commit` allows integration of `selene` into your Git workflow using git hooks. After [installing pre-commit](https://pre-commit.com/#install), add one of the following configurations to your `.pre-commit-config.yaml` file:
+
+* Use the `selene` binary present on the system path (Should be pre-installed):
+
+```yaml
+repos:
+  - repo: https://github.com/Kampfkarren/selene
+    rev: ''
+    hooks:
+      - id: selene-system
+```
+
+* Use `selene` through GitHub releases:
+
+```yaml
+repos:
+  - repo: https://github.com/Kampfkarren/selene
+    rev: ''
+    hooks:
+      - id: selene-github
+```
+
+* Use the `selene` binary present in the `selene` docker image (Since this uses docker, it might take some time to bootstrap and is slower than the other two options):
+
+```yaml
+repos:
+  - repo: https://github.com/Kampfkarren/selene
+    rev: ''
+    hooks:
+      - id: selene-docker
+```
+
+You may see a `warning` being generated when pre-commit runs. To resolve that, set the `rev` key to any selene tag or commit, for e.g. `rev: '0.26.2'`.
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,9 @@
+[build-system]
+requires = ["release-gitter[builder]"]
+build-backend = "pseudo_builder"
+
+[tool.release-gitter]
+git-url = "https://github.com/Kampfkarren/selene"
+extract-files = ["selene"]
+format = "selene-{version}-{system}.zip"
+exec = "chmod +x selene"
--- a/Dockerfile
+++ b/Dockerfile
@@ -28,10 +28,10 @@ FROM bash AS selene-light
 COPY --from=selene-light-builder /usr/local/cargo/bin/selene /
 CMD ["/selene"]
 
-FROM bash AS selene-musl
-COPY --from=selene-musl-builder /usr/local/cargo/bin/selene /
-CMD ["/selene"]
-
 FROM bash AS selene-light-musl
 COPY --from=selene-light-musl-builder /usr/local/cargo/bin/selene /
+CMD ["/selene"]
+
+FROM bash AS selene-musl
+COPY --from=selene-musl-builder /usr/local/cargo/bin/selene /
 CMD ["/selene"]


Cargo.toml
```diff
@@ -1,5 +1,5 @@
 [workspace]
-
+resolver = "2"
```
amitds1997, line 2: Without adding this, `cargo install` throws the following warnings:
```text
warning: some crates are on edition 2021 which defaults to `resolver = "2"`, but virtual workspaces default to `resolver = "1"`
note: to keep the current resolver, specify `workspace.resolver = "1"` in the workspace root's manifest
note: to use the edition 2021 resolver, specify `workspace.resolver = "2"` in the workspace root's manifest
warning: some crates are on edition 2021 which defaults to `resolver = "2"`, but virtual workspaces default to `resolver = "1"`
note: to keep the current resolver, specify `workspace.resolver = "1"` in the workspace root's manifest
note: to use the edition 2021 resolver, specify `workspace.resolver = "2"` in the workspace root's manifest
```
Set the resolver to "2" based on [rust workspaces docs](https://doc.rust-lang.org/cargo/reference/workspaces.html#virtual-workspace).

.pre-commit-hooks.yaml
```diff
@@ -0,0 +1,15 @@
+- id: selene-system
```
amitds1997, line 1: Rust-based pre-commit does not work because pre-commit hardcodes the `--path` to `.` ([source](https://github.com/pre-commit/pre-commit/blob/14169eb31db6a0ac3e4dfbc38234a8a3e90bc6c1/pre_commit/languages/rust.py#L135)) which leads to build failure with error: `found a virtual manifest at <path>/Cargo.toml instead of a package manifest`. 

Pre-commit maintainer disagrees with making `--path` configurable so it is not possible to add a rust based hook: https://github.com/pre-commit/pre-commit/issues/2931 unless we somehow make it possible to install the binary from root level.

chriscerie (changes requested):
Need changelog

docs/src/SUMMARY.md
```diff
@@ -10,6 +10,7 @@
   - [Configuration](./usage/configuration.md)
   - [Filtering](./usage/filtering.md)
   - [Standard Library Format](./usage/std.md)
+  - [Version control integration](./usage/version-control-integration.md)
```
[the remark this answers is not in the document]
amitds1997: Let me update that to "Git hooks (pre-commit)"

.pre-commit-hooks.yaml
amitds1997: I was unsure if it was okay to add a custom install script only for the pre-commit use case. I can add it if that's something we are okay with.

docs/src/SUMMARY.md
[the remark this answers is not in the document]
amitds1997: Done in [6391ba6](https://github.com/Kampfkarren/selene/pull/565/commits/6391ba6affbab50c80deff79efffa1ef7d2b011e)

.pre-commit-hooks.yaml
chriscerie: I think it would depend on how it would look like. Caching binaries by version may not be trivial and we don't get the same advantages `language: rust` gets, which makes me think it's likely not worth the effort to support this.

amitds1997: Agree. That's why went ahead with only setting up `system` and `docker` workflows. Is there some trivial way to make this work by may be a configuration change at our end? I did some preliminary research but did not find much (but then I have not worked with `rust` ever before).

chriscerie (commented):
It looks like pre-commit builds docker with `docker build .`, which causes the light build to be used. This isn't compatible with roblox projects. It also doesn't seem like pre-commit supports changing the build command, so a rather hacky workaround could be to move the `selene-musl` build to the last in the dockerfile.

.pre-commit-hooks.yaml
jenstroeger: Have you looked at how [StyLua handles the pre-commit hook](https://github.com/JohnnyMorganz/StyLua?#pre-commit)?

amitds1997: Their `rust` workflow works because they do not use workspaces.

The `python` workflow seems interesting. We can integrate it, but it would depend on a python package (https://pypi.org/project/release-gitter/) to do the release download for us. If that is something we are okay with, I can add it 👍

CHANGELOG.md
```diff
@@ -1,7 +1,10 @@
+<<<<<<< HEAD
```
chriscerie, line 1: Looks like merging has gone wrong.

docs/src/usage/git-hooks-pre-commit.md
```diff
@@ -0,0 +1,10 @@
+To integrate selene with git, you can use [pre-commit](https://pre-commit.com). Once you have [pre-commit installed](https://pre-commit.com/#install), add this to the `.pre-commit-config.yaml` at root of your repository:
+```yaml
+repos:
+  - repo: https://github.com/Kampfkarren/selene
+    rev: '' # Change this value to use a different version of Selene
```
chriscerie, line 5: Does this do anything with `selene-system`, since it uses whatever is installed locally? If not then I think it'd be less confusing to separate the system and docker examples.

CHANGELOG.md
amitds1997: Fixed in [5f1a755](https://github.com/Kampfkarren/selene/pull/565/commits/5f1a755b32acc8d36ed96c417eabaf81159018ee)

docs/src/usage/git-hooks-pre-commit.md
amitds1997: Separated out each type of hook into it's own piece of documentation in [5f1a755](https://github.com/Kampfkarren/selene/pull/565/commits/5f1a755b32acc8d36ed96c417eabaf81159018ee)

.pre-commit-hooks.yaml
amitds1997: Added GitHub release based pre-commit hook in [5f1a755](https://github.com/Kampfkarren/selene/pull/565/commits/5f1a755b32acc8d36ed96c417eabaf81159018ee)

amitds1997:
@chriscerie @jenstroeger Let me know if there's something else needed from my end to get this through.

chriscerie:
I’m currently ooo, I’ll take a look at the changes within the next week or so.

chriscerie:
@amitds1997 Are you able to upload the builds under your fork's release? I think that's needed to test selene-github.

amitds1997:
> @amitds1997 Are you able to upload the builds under your fork's release? I think that's needed to test selene-github.

I was able to test it out with this:
```yaml
  - repo: https://github.com/amitds1997/selene
    rev: d895ab518a9a3fd6e8efe1716fde501dd29fe8ba # The latest commit on the PR branch
    hooks:
      - id: selene-github
```
Did not face any issues. The reason I think this should work is that it looks for releases under Kampfkarren/selene's releases (see [amitds1997/selene/pyproject.toml#L6](https://github.com/amitds1997/selene/blob/d895ab518a9a3fd6e8efe1716fde501dd29fe8ba/pyproject.toml#L6)). Are you seeing any particular error?

chriscerie:
> Are you seeing any particular error?

It works when I have a local copy of selene, but it's breaking when it's not installed:
```
selene (GitHub)..........................................................Failed
- hook id: selene-github
- exit code: 1

Executable `selene` not found
```

You're right about it using git-url to get the upstream repo. Perhaps where it's breaking is it's missing `extract-files` (https://github.com/JohnnyMorganz/StyLua/blob/main/pyproject.toml#L7) and it's falling back to a local copy of selene if installed?

IamTheFij:
Hello. I just started looking into Selene and noticed that this is kind of waiting on me! Sorry about that. I just pushed a new version that should resolve this issue and allow the GitHub install hook to properly download.

amitds1997:
@chriscerie @jenstroeger With the latest release to [release-gitter](https://github.com/IamTheFij/release-gitter) (thanks @IamTheFij), we have a working solution for pip-based pre-commit installation.

With this, we should have pre-commit support using local exec, docker and pip (python). Let me know if anything else would be needed from my end.

pyproject.toml
```diff
@@ -0,0 +1,14 @@
...
+[tool.release-gitter]
+git-url = "https://github.com/Kampfkarren/selene"
+extract-files = ["selene"]
+format = "selene-{version}-{system}.zip"
+exec = "chmod +x selene"
+
+[tool.release-gitter.map-system]
+Darwin = "macos"
+Windows = "windows"
+Linux = "linux"
```
IamTheFij, line 14: These should no longer be needed. This was common enough that I added it as default values. That said, there is no harm in leaving them.

https://github.com/IamTheFij/release-gitter/blob/main/release_gitter.py#L51
```python
SYSTEM_SYNONYMS: list[list[str]] = [
    ["Darwin", "darwin", "MacOS", "macos", "macOS"],
    ["Windows", "windows", "win", "win32", "win64"],
    ["Linux", "linux"],
]
```

amitds1997: Removed this in the latest commit ([391b18d](https://github.com/Kampfkarren/selene/pull/565/commits/391b18dead24fa6bfc6f446f814a91d4edfa6e0f)).

IamTheFij (approved):
Looks good!

amitds1997:
> @amitds1997 any help needed here

I think this just fell through the cracks. I don't think there was anything pending from my end. If the maintainers want this in, I can do a cleanup and get this ready again.

benner:
Me too

PR #574, opened 2023-12-02
Lint against approximated constants

### LINKED ISSUE #573: Lint against approximated constants
Should favor constants like `math.pi` over `3.14`. It should also work for both truncated and rounded numbers, so `3.141` and `3.142` should both lint. But `3.143` should not. It should also work for any arbitrary precision.

chriscerie:
Closes #573.


Kampfkarren (changes requested):
Code all looks fine, but can we fully name it "approximated_constants"?

PR #578, opened 2024-01-13
Fix vscode extension not respecting exclude

chriscerie:


Kampfkarren (commented):
Does this break selene if it's in a new tab with no associated file?

PR #579, opened 2024-01-14
Update sct and ureq to bump ring>=0.17

klemensn:
On OpenBSD, ring 0.16 a) needs patches to comply with execute-only memory policies and b) does not build on sparc64, riscv64 and powerpc64.

Done with `cargo ureq sct`.


chriscerie (approved):
Thanks!

Issue #407, opened 2022-08-28
Standard library format docs don't mention must_use


Kampfkarren:
https://kampfkarren.github.io/selene/usage/std.html

You can add `must_use` to globals to trigger the must_use lint.

PR #583, opened 2024-01-24
Add `must_use` to std docs

Midnightific:
Closes: #407 

Adds `must_use` functionality and documentation for std docs.

### DIFF
--- a/docs/src/usage/std.md
+++ b/docs/src/usage/std.md
@@ -67,6 +67,17 @@ This field is used for allowing smarter introspection of how the argument given 
 - "read" - This argument is only read from. Currently unused.
 - "write" - This argument is only written to. Used by `unused_variable` to assist in detecting a variable only being written to, even if passed into a function.
 
+Example:
+```yml
+ table.insert:
+  args:
+    - type: table
+      observes: write # This way, `table.insert(x, 1)` doesn't count as a read to `x`
+    - type: any
+    - required: false
+      type: any
+```
+
 #### "must_use"
 This field is used for checking if the return value of a function is used.
 
@@ -78,10 +89,10 @@ Example:
 table.find:
   args:
     - type: table
-      observes: read  # The table is only read from
-      required: true  # The table to index is a required argument
+      observes: read
+      required: true
     - type: any
-      required: true  # The value to find is a required argument
+      required: true
   must_use: true  # The return value of this function must be used
 ```
 
--- a/docs/src/usage/std.md
+++ b/docs/src/usage/std.md
@@ -67,15 +67,22 @@ This field is used for allowing smarter introspection of how the argument given 
 - "read" - This argument is only read from. Currently unused.
 - "write" - This argument is only written to. Used by `unused_variable` to assist in detecting a variable only being written to, even if passed into a function.
 
+#### "must_use"
+This field is used for checking if the return value of a function is used.
+
+- `true` - The default. The return value of this function must be used.
+- `false` - The return value of this function does not need to be used.
+
 Example:
 ```yml
- table.insert:
+table.find:
   args:
     - type: table
-      observes: write # This way, `table.insert(x, 1)` doesn't count as a read to `x`
+      observes: read  # The table is only read from
+      required: true  # The table to index is a required argument
     - type: any
-    - required: false
-      type: any
+      required: true  # The value to find is a required argument
+  must_use: true  # The return value of this function must be used
 ```
 
 #### Argument types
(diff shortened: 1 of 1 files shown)


docs/src/usage/std.md
```diff
@@ -67,15 +67,22 @@ This field is used for allowing smarter introspection of how the argument given
 - "read" - This argument is only read from. Currently unused.
 - "write" - This argument is only written to. Used by `unused_variable` to assist in detecting a variable only being written to, even if passed into a function.
 
+#### "must_use"
+This field is used for checking if the return value of a function is used.
+
+- `true` - The default. The return value of this function must be used.
```
chriscerie, line 73: The default is false.

docs/src/usage/std.md
```diff
@@ -67,15 +67,22 @@ This field is used for allowing smarter introspection of how the argument given
...
+- `false` - The return value of this function does not need to be used.
+
 Example:
 ```yml
- table.insert:
```
chriscerie, line 72: Why remove the example for `observes`?

docs/src/usage/std.md
```diff
@@ -67,15 +67,22 @@ This field is used for allowing smarter introspection of how the argument given
...
+table.find:
   args:
     - type: table
-      observes: write # This way, `table.insert(x, 1)` doesn't count as a read to `x`
+      observes: read  # The table is only read from
+      required: true  # The table to index is a required argument
     - type: any
-    - required: false
-      type: any
+      required: true  # The value to find is a required argument
```
chriscerie, line 84: This is a section for `must_use`, not `required`. This comment isn't needed. Same with the others.

chriscerie (changes requested):

docs/src/usage/std.md
Midnightific: Could I just give a example for each one instead? I can't think of a native function of Luau that satifies all these.

docs/src/usage/std.md
Midnightific: See https://github.com/Kampfkarren/selene/pull/583#discussion_r1465686806 for what I was attempting to do.

docs/src/usage/std.md
chriscerie: Yes that’s what I mean. In your diff you removed the example for the previous section. Keep the previous one and make a new example for `must_use`.

docs/src/usage/std.md
Midnightific: Resolved in: https://github.com/Kampfkarren/selene/pull/583/commits/b7c23500aa029dff2ea99af1d2c18af418bcd4de

docs/src/usage/std.md
Midnightific: Resolved in: https://github.com/Kampfkarren/selene/pull/583/commits/607ca211d56b7850df50006631eceb5ebc9c87a6

docs/src/usage/std.md
Midnightific: Resolved in: https://github.com/Kampfkarren/selene/pull/583/commits/607ca211d56b7850df50006631eceb5ebc9c87a6

docs/src/usage/std.md
```diff
@@ -78,6 +78,24 @@ Example:
       type: any
 ```
 
+#### "must_use"
+This field is used for checking if the return value of a function is used.
+
+- `true` - The return value of this function must be used.
+- `false` - The default. The return value of this function does not need to be used.
```
chriscerie, line 85: Default option should come first.

docs/src/usage/std.md
```diff
@@ -78,6 +78,24 @@ Example:
...
+Example:
+```yml
+table.find:
+  args:
+    - type: table
+      observes: read
+      required: true
+    - type: any
+      required: true
+  must_use: true  # The return value of this function must be used
```
chriscerie, line 96: See the existing spec for `table.find`. Let's keep the example consistent and correct.
https://github.com/Kampfkarren/selene/blob/65174d1f056ce71442a34a9aeb9ed9c968c82b96/selene-lib/default_std/luau.yml#L376-L382

chriscerie (changes requested):

docs/src/usage/std.md
Midnightific: Resolved in: https://github.com/Kampfkarren/selene/pull/583/commits/78b1523b8d906fbc608a977b9c17d6d03bc1d72e

docs/src/usage/std.md
Midnightific: Resolved in: https://github.com/Kampfkarren/selene/pull/583/commits/9dd838a2143095670586564e7c15a5434578eff4

Midnightific:
@Kampfkarren Are you able to take a look at this?

Issue #572, opened 2023-11-29
Better error message for mixed tables


chriscerie:
"mixed tables are not allowed" doesn't actually say why it's bad. It's also easy to misinterpret this since mixed tables are valid, but are just heavily discouraged.
```
error[mixed_table]: mixed tables are not allowed
  ┌─ mixed_table.lua:2:5
  │  
2 │ ╭     "",
3 │ │     a = b,
  │ ╰─────────^
  │  
  = help: change this table to either an array or dictionary
  ```

PR #586, opened 2024-02-05
Error mixed table

Midnightific:
Closes: #572 

It makes the error message for mixed tables more descriptive and explains why they're bad and should be _avoided_.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 ## [Unreleased](https://github.com/Kampfkarren/selene/compare/0.26.1...HEAD)
 ### Added
 - Added `CFrame.lookAlong` to the Roblox standard library
+- When a mixed table is detected, selene will now tell you why mixed tables are bad and that they should be avoided
 
 ## [0.26.1](https://github.com/Kampfkarren/selene/releases/tag/0.26.1) - 2023-11-11
 ### Fixed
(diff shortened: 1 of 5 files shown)


selene-lib/src/lints/mixed_table.rs
```diff
@@ -30,7 +30,7 @@ impl Lint for MixedTableLint {
         for mixed_table in visitor.mixed_tables {
             diagnostics.push(Diagnostic::new_complete(
                 "mixed_table",
-                "mixed tables are not allowed".to_owned(),
+                "mixed tables are heavily discouraged due to factors such as: type safety, readability, compatibility, and debugging".to_owned(),
```
Kampfkarren, line 33: The message here does not really adequately explain why they should be avoided. Also, "type safety" is not a thing outside of Luau. I don't think it's ever possible to do that in a small message, so I suggest something more openly vague:

```suggestion
                "mixed tables should be avoided, as they can cause confusing and hard to debug issues such as during iteration or encoding".to_owned(),
```

Midnightific: Aye aye! *Sorry I forgot that selene isn't a linter for just Roblox*

Midnightific:
Scrap that review, forgot to change the stderr file, will do that tomorrow.

Midnightific:
@Kampfkarren All good to review now!

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 ## [Unreleased](https://github.com/Kampfkarren/selene/compare/0.26.1...HEAD)
 ### Added
 - Added `CFrame.lookAlong` to the Roblox standard library
+- When a mixed table is detected, selene will now tell you why mixed tables are bad and that they should be avoided
```
chriscerie, line 7:
```suggestion

### Changed
- Updated the warning message for the `must_use` lint to include why mixed tables should be avoided
```

Issue #469, opened 2022-12-08
Selene should search for configuration files that begin with a dot


tomodachi94:
Selene should allow prefixing all of its configuration with a dot. [StyLua allows this.](https://github.com/JohnnyMorganz/StyLua#finding-the-configuration)

This would aid with having developer tooling configurations being distinguished from ordinary source files.

tomodachi94:
For example, I would have a file, `.selene.toml`, that contains the same things as `selene.toml`. In addition, I would have a `.my-custom-stdlib.yaml` file that contains my custom standard library.

tomodachi94:
Workaround: Specify the `--config` option when running `selene`.

Kampfkarren:
`.selene.toml` seems fine, but I think standard libraries should just work with putting the dot in the std name itself, we already have a bunch of weird combinatorics with them

Zeioth:
This is a convention of most formaters/linters. It would be very desireable.

chriscerie:
Is there a stronger motivation for this? Introducing multiple ways to do the same thing is not great, and eslint is [moving away from such behavior](https://eslint.org/blog/2023/10/flat-config-rollout-plans/#eslintrc-removed-in-eslint-v10.0.0).

LastTalon:
> eslint is [moving away from such behavior](https://eslint.org/blog/2023/10/flat-config-rollout-plans/#eslintrc-removed-in-eslint-v10.0.0).

ESLint is moving away from their old config format entirely. It's not really about prefixing with a `.`, which is usually just to treat the config as a hidden file.

chriscerie:
> > eslint is [moving away from such behavior](https://eslint.org/blog/2023/10/flat-config-rollout-plans/#eslintrc-removed-in-eslint-v10.0.0).
> 
> ESLint is moving away from their old config format entirely. It's not really about prefixing with a `.`, which is usually just to treat the config as a hidden file.

I more meant there doesn’t seem to be a strong enough use case for eslint to support dot-prefixed configs in their new system. What’s different about lua where this would be worth a more complicated config resolution?

chriscerie:
I'm going to close this for now as I don't see the benefits worth the added complexity. I'm not entirely opposed to reopening this in the future if there's a stronger motivation.

Zeioth:
In case you are not familiar, on UNIX system (Linux, MacOS, and others), adding a dot in front of the file, make the file a hidden for file browsers. That way we avoid polluting our user projects.

[...]

PR #591, opened 2024-03-10
✨feat: `.selene.toml` and `selene.toml` are now detected by order of priority

Zeioth:
This PR:
* add 20 lines of code, but it keep all the original checkings.
* I've tried it to look as similar to the original code as possible. 
* It should be obvious enough.

The feature has been tested on neovim + none-ls with:
* No file.
* .selene.toml
* selene.toml
* .selene.toml + wrong syntax
* selene.toml + wrong syntax

Example of bad syntax
screenshot_2024-03-10_04-17-20_778230898

Example of file detected OK
screenshot_2024-03-10_04-18-01_912823820

Example of no config file
screenshot_2024-03-10_04-19-59_927521326

### DIFF
--- a/docs/src/cli/usage.md
+++ b/docs/src/cli/usage.md
@@ -16,7 +16,7 @@ FLAGS:
 
 OPTIONS:
         --color <color>                     [default: auto]  [possible values: Always, Auto, Never]
-        --config <config>                  A toml file to configure the behavior of selene [default: selene.toml]
+        --config <config>                  A toml file to configure the behavior of selene [default: .selene.toml]
         --display-style <display-style>    Sets the display method [possible values: Json, Json2, Rich, Quiet]
         --num-threads <num-threads>        Number of threads to run on, default to the numbers of logical cores on your
                                            system [default: your system's cores]
--- a/selene/src/main.rs
+++ b/selene/src/main.rs
@@ -412,7 +412,7 @@ fn start(mut options: opts::Options) {
 
                 (config_contents, Path::new("-"))
             } else {
-                let config_paths_to_check = [Path::new("selene.toml"), Path::new(".selene.toml")];
+                let config_paths_to_check = [Path::new(".selene.toml"), Path::new("selene.toml")];
                 let mut config_contents = String::new();
                 let mut config_path = Path::new("");
                 let mut config_not_found = true;
@@ -541,8 +541,8 @@ fn start(mut options: opts::Options) {
                         .ok()
                 };
 
-                let config = read_config_file("selene.toml")
-                    .or_else(|| read_config_file(".selene.toml"))
+                let config = read_config_file(".selene.toml")
+                    .or_else(|| read_config_file("selene.toml"))
                     .unwrap_or_else(|| CheckerConfig::default());
 
                 (config, None)
--- a/selene/src/main.rs
+++ b/selene/src/main.rs
@@ -391,7 +391,39 @@ fn read_file(checker: &Checker<toml::value::Value>, filename: &Path) {
     );
 }
 
-fn read_config_file() -> Option<(CheckerConfig<toml::Value>, Option<PathBuf>)> {
+fn read_config_file_as_string() -> Option<(String, &'static Path)> {
+    let config_paths_to_check = [".selene.toml", "selene.toml"];
+    let mut config_contents = String::new();
+    let mut config_path = None;
+
+    for path in &config_paths_to_check {
+        if let Ok(contents) = fs::read_to_string(path) {
+            config_contents = contents;
+            config_path = Some(PathBuf::from(path)); // Convert str to PathBuf
+            break;
+        }
+    }
+
+    if config_contents.is_empty() {
+        return None;
+    }
+
+    // Create a static variable to hold config_path
+    static mut CONFIG_PATH: Option<PathBuf> = None;
+
+    // If config_path is Some, assign it to the static variable
+    if let Some(path) = config_path {
+        unsafe {
+            CONFIG_PATH = Some(path);
+        }
+    }
+
+    // Parse config_contents into String
+    Some((config_contents, unsafe { CONFIG_PATH.as_ref().unwrap().as_path() }))
+}
+
+
+fn read_config_file_as_checkerconfig() -> Option<(CheckerConfig<toml::Value>, Option<PathBuf>)> {
     let config_paths_to_check = [".selene.toml", "selene.toml"];
     let mut config_contents = String::new();
     let mut config_path = None;
@@ -440,42 +472,15 @@ fn start(mut options: opts::Options) {
 
                 (config_contents, Path::new("-"))
             } else {
+                let (config_content, config_path) = match read_config_file_as_string() {
+                    Some((config_content, config_path)) => (config_content, config_path),
+                    None => {
+                      error!("Error reading config file");
+                      std::process::exit(1);
+                    },
+                };
 
-                // How it should look
-                // let (config, config_path) = match read_config_file() {
-                //     Some((config, config_path)) => (config, config_path),
-                //     None => {
-                //       error!("Error reading config file");
-                //       std::process::exit(1);
-                //     },
-                // };
-                //
-                // (config, config_path)
-
-
-                // validate config
-                let config_paths_to_check = [Path::new(".selene.toml"), Path::new("selene.toml")];
-                let mut config_contents = String::new();
-                let mut config_path = Path::new("");
-                let mut config_not_found = true;
-
-                for path in &config_paths_to_check {
-                    if path.exists() {
-                        if let Ok(contents) = fs::read_to_string(path) {
-                            config_contents = contents;
-                            config_path = path;
-                            config_not_found = false;
-                            break;
-                        }
-                    }
-                }
-
-                if config_not_found {
-                    error!("Error reading config file");
-                    std::process::exit(1);
-                }
-
-                (config_contents, config_path)
+                (config_content, config_path)
             };
 
             if let Err(error) = validate_config::validate_config(
@@ -571,7 +576,7 @@ fn start(mut options: opts::Options) {
             }
 
             None => {
-                let (config, _) = match read_config_file() {
+                let (config, _) = match read_config_file_as_checkerconfig() {
                     Some(config) => config,
                     None => (CheckerConfig::default(), None),
                 };
--- a/selene/src/main.rs
+++ b/selene/src/main.rs
@@ -396,6 +396,7 @@ fn read_config_file(file: &str) -> Option<CheckerConfig<toml::value::Value>> {
     toml::from_str(&config_contents)
         .map_err(|error| {
             error!("Error parsing config file {file}: {error}");
+            std::process::exit(1);
         })
         .ok()
 }
--- a/selene/src/main.rs
+++ b/selene/src/main.rs
@@ -391,6 +391,15 @@ fn read_file(checker: &Checker<toml::value::Value>, filename: &Path) {
     );
 }
 
+fn read_config_file(file: &str) -> Option<CheckerConfig<toml::value::Value>> {
+    let config_contents = fs::read_to_string(file).ok()?;
+    toml::from_str(&config_contents)
+        .map_err(|error| {
+            error!("Error parsing config file {}: {}", file, error);
+        })
+        .ok()
+}
+
 fn start(mut options: opts::Options) {
     *OPTIONS.write().unwrap() = Some(options.clone());
 
@@ -531,15 +540,6 @@ fn start(mut options: opts::Options) {
             }
 
             None => {
-                let read_config_file = |file: &str| -> Option<CheckerConfig<toml::value::Value>> {
-                    let config_contents = fs::read_to_string(file).ok()?;
-                    toml::from_str(&config_contents)
-                        .map_err(|error| {
-                            error!("Error parsing config file {}: {}", file, error);
-                        })
-                        .ok()
-                };
-
                 let config = read_config_file(".selene.toml")
                     .or_else(|| read_config_file("selene.toml"))
                     .unwrap_or_else(|| CheckerConfig::default());
--- a/selene/src/main.rs
+++ b/selene/src/main.rs
@@ -412,15 +412,28 @@ fn start(mut options: opts::Options) {
 
                 (config_contents, Path::new("-"))
             } else {
-                let config_path = Path::new("selene.toml");
+                let config_paths_to_check = [Path::new("selene.toml"), Path::new(".selene.toml")];
+                let mut config_contents = String::new();
+                let mut config_path = Path::new("");
+                let mut config_not_found = true;
 
-                let config_contents = match fs::read_to_string(config_path) {
-                    Ok(contents) => contents,
-                    Err(error) => {
-                        error!("Error reading config file: {error}");
-                        std::process::exit(1);
+                for path in &config_paths_to_check {
+                    match fs::read_to_string(path) {
+                        Ok(contents) => {
+                            config_contents = contents;
+                            config_path = path;
+                            config_not_found = false;
+                            break;
+                        }
+                        Err(error) => {
+                            error!("Error reading config file: {error}");
+                        }
                     }
-                };
+                }
+
+                if config_not_found {
+                    std::process::exit(1);
+                }
 
                 (config_contents, config_path)
             };
@@ -517,17 +530,23 @@ fn start(mut options: opts::Options) {
                 }
             }
 
-            None => match fs::read_to_string("selene.toml") {
-                Ok(config_contents) => match toml::from_str(&config_contents) {
-                    Ok(config) => (config, None),
-                    Err(error) => {
-                        error!("Config file not in correct format: {}", error);
-                        std::process::exit(1);
-                    }
-                },
+            None => {
+                let read_config_file = |file: &str| -> Option<CheckerConfig<toml::value::Value>> {
+                    let config_contents = fs::read_to_string(file).ok()?;
+                    toml::from_str(&config_contents)
+                        .map_err(|error| {
+                            eprintln!("Error parsing config file {}: {}", file, error);
+                            error
+                        })
+                        .ok()
+                };
 
-                Err(_) => (CheckerConfig::default(), None),
-            },
+                let config = read_config_file("selene.toml")
+                    .or_else(|| read_config_file(".selene.toml"))
+                    .unwrap_or_else(|| CheckerConfig::default());
+
+                (config, None)
+            }
         };
 
     let current_dir = std::env::current_dir().unwrap();
--- a/selene/src/main.rs
+++ b/selene/src/main.rs
@@ -535,8 +535,7 @@ fn start(mut options: opts::Options) {
                     let config_contents = fs::read_to_string(file).ok()?;
                     toml::from_str(&config_contents)
                         .map_err(|error| {
-                            eprintln!("Error parsing config file {}: {}", file, error);
-                            error
+                            error!("Error parsing config file {}: {}", file, error);
                         })
                         .ok()
                 };
--- a/selene/src/main.rs
+++ b/selene/src/main.rs
@@ -428,20 +428,18 @@ fn start(mut options: opts::Options) {
                 let mut config_not_found = true;
 
                 for path in &config_paths_to_check {
-                    match fs::read_to_string(path) {
-                        Ok(contents) => {
+                    if path.exists() {
+                        if let Ok(contents) = fs::read_to_string(path) {
                             config_contents = contents;
                             config_path = path;
                             config_not_found = false;
                             break;
                         }
-                        Err(error) => {
-                            error!("Error reading config file: {error}");
-                        }
                     }
                 }
 
                 if config_not_found {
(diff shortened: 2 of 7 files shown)


Zeioth:
Also it closes #469

selene/src/main.rs
```diff
@@ -412,15 +412,28 @@ fn start(mut options: opts::Options) {
...
+                        Ok(contents) => {
+                            config_contents = contents;
+                            config_path = path;
+                            config_not_found = false;
+                            break;
+                        }
+                        Err(error) => {
+                            error!("Error reading config file: {error}");
+                        }
                     }
```
chriscerie, line 431: This wouldn't work because you're propagating `fs::read_to_string` errors even when the path doesn't exist. You'd need to handle that separately:

```suggestion
                [".selene.toml", "selene.toml"]
                    .iter()
                    .find_map(|path| {
                        let path = Path::new(path);
                        path.exists().then_some(path).and_then(|path| {
                            fs::read_to_string(path).map_or_else(
                                |e| {
                                    error!("Error reading config file: {e}");
                                    std::process::exit(1);
                                },
                                |contents| Some((contents, path)),
                            )
                        })
                    })
                    .unwrap_or_else(|| {
                        error!("No config file found");
                        std::process::exit(1);
                    })
```

selene/src/main.rs
```diff
@@ -412,15 +412,28 @@ fn start(mut options: opts::Options) {
...
-                };
+                }
+
+                if config_not_found {
+                    std::process::exit(1);
+                }
 
                 (config_contents, config_path)
```
chriscerie, line 438: See comment above.
```suggestion
```

selene/src/main.rs
```diff
@@ -517,17 +530,23 @@ fn start(mut options: opts::Options) {
...
+                            eprintln!("Error parsing config file {}: {}", file, error);
+                            error
+                        })
+                        .ok()
+                };
 
-                Err(_) => (CheckerConfig::default(), None),
-            },
+                let config = read_config_file("selene.toml")
+                    .or_else(|| read_config_file(".selene.toml"))
```
chriscerie, line 545: Dotfiles should take precedence.

selene/src/main.rs
```diff
@@ -517,17 +530,23 @@ fn start(mut options: opts::Options) {
...
-                        error!("Config file not in correct format: {}", error);
-                        std::process::exit(1);
-                    }
-                },
+            None => {
+                let read_config_file = |file: &str| -> Option<CheckerConfig<toml::value::Value>> {
+                    let config_contents = fs::read_to_string(file).ok()?;
+                    toml::from_str(&config_contents)
+                        .map_err(|error| {
+                            eprintln!("Error parsing config file {}: {}", file, error);
```
chriscerie, line 538: Use `error!` for consistent formatting.

selene/src/main.rs
```diff
@@ -517,17 +530,23 @@ fn start(mut options: opts::Options) {
...
+                            error
+                        })
+                        .ok()
```
chriscerie, line 541: Need to early terminate the process when config file is invalid.

selene/src/main.rs
```diff
@@ -517,17 +530,23 @@ fn start(mut options: opts::Options) {
...
-            None => match fs::read_to_string("selene.toml") {
-                Ok(config_contents) => match toml::from_str(&config_contents) {
-                    Ok(config) => (config, None),
-                    Err(error) => {
-                        error!("Config file not in correct format: {}", error);
-                        std::process::exit(1);
-                    }
-                },
+            None => {
+                let read_config_file = |file: &str| -> Option<CheckerConfig<toml::value::Value>> {
```
chriscerie, line 534: Since finding the config file isn't so trivial anymore we should move the one from validate config command to a separate function and use that instead.

chriscerie (changes requested):
Need changelog + docs.

Zeioth:
@chriscerie I'm gonna try but this is my first contribution to a rust project and I'm not a expert in the language so I feelI might cause more noise than it's worth trying to apply the feedback.

If you want to co-author please go ahead.

selene/src/main.rs
Zeioth: fixed in https://github.com/Kampfkarren/selene/pull/591/commits/02213f435c44dfb46e66bb7e43a2aa52bd48ef9a

selene/src/main.rs
Zeioth: solved in https://github.com/Zeioth/selene/commit/b4a93e328ebbdefcd336278648952b0974610117

selene/src/main.rs
Zeioth: refactored in https://github.com/Zeioth/selene/commit/13e628625f41b99d38aa817fd9c94a33a2e8b1ed

## More info
The code to read and validate a config file is now a separated function.
```rust
fn read_config_file(file: &str) -> Option<CheckerConfig<toml::value::Value>> {
    let config_contents = fs::read_to_string(file).ok()?;
    toml::from_str(&config_contents)
        .map_err(|error| {
            error!("Error parsing config file {}: {}", file, error);
        })
        .ok()
}
```

chriscerie: I meant more like a separate function to find the config file. It’s needed in the validate config command logic as well: https://github.com/Kampfkarren/selene/pull/591#discussion_r1519041670

selene/src/main.rs
Zeioth: Closed on https://github.com/Zeioth/selene/commit/de9aa798418e98ee77f8b3d91e8ee865ff66fc49

selene/src/main.rs
Zeioth: fixed in https://github.com/Zeioth/selene/commit/8947dea186d45ba65b5e8196b670bb6750ef83b8

selene/src/main.rs
Zeioth: fixed in https://github.com/Zeioth/selene/commit/8947dea186d45ba65b5e8196b670bb6750ef83b8

selene/src/main.rs
Zeioth: @chriscerie ok this should do it:

* All logic regarding reading and parsing config files has been moved to dedicated functions:

```rust
/// Reads a config file and return content and path. None if not found or empty.
fn read_config_file() -> Option<(String, Option<PathBuf>)> {
    let config_paths_to_check = [".selene.toml", "selene.toml"];
    let mut config_contents = String::new();
    let mut config_path = None;

    for path in &config_paths_to_check {
        if let Ok(contents) = fs::read_to_string(path) {
            config_contents = contents;
            config_path = Some(PathBuf::from(path));
            break;
        }
    }

    if config_contents.is_empty() {
        return None;
    }

    Some((config_contents, config_path))
}

/// Read the config file and return it as Option<(String, &'static Path)>.
fn parse_config_file_as_string() -> Option<(String, &'static Path)> {
    if let Some((config_contents, config_path)) = read_config_file() {
        if let Some(path) = config_path {
            return Some((config_contents, Box::leak(Box::new(path)).as_path()));
        }
    } else {
        error!("Error reading config file");
        std::process::exit(1);
    }
    None
}

/// Read the config file and return it as CheckerConfig<toml::Value>, Option<PathBuf>.
fn parse_config_file_as_checkerconfig() -> Option<(CheckerConfig<toml::Value>, Option<PathBuf>)> {
    if let Some((config_contents, config_path)) = read_config_file() {
        match toml::from_str(&config_contents) {
            Ok(config) => Some((config, config_path)),
            Err(error) => {
                error!("Error parsing config file: {}", error);
                std::process::exit(1);
            }
        }
    } else {
        None
    }
}
```

The calls now look like:
```
let (config_content, config_path) = parse_config_file_as_string().unwrap();
(config_content, config_path)
```
and
```
let (config, _) = match parse_config_file_as_checkerconfig() {
    Some(config) => config,
    None => (CheckerConfig::default(), None),
};

(config, None)
```

The errors are now controlled inside the functions because it was the only way to keep our current logic intact (and it's probably cleaner too).

Zeioth:
> Need changelog + docs.

Changelog: https://github.com/Zeioth/selene/blob/dot-selene-toml/CHANGELOG.md
Docs: https://github.com/Zeioth/selene/commit/e5fd4f0fab47fbf515c7d60aaca30eeb7c50bb95

## More info
In the docs, I've just replaced every mention to `selene.toml` by `.selene.toml` since now `.selene.toml` has priority over `selene.toml`.

Only in `src/usage/configuration` we mention that we also accept `selene.toml` to avoid duplication.

Zeioth:
The PR should be ready merge it. Please tell me if you would like some other changes.

Hopefully we can ship it relatively soon.

chriscerie:
Thanks for working on this! Unfortunately I've closed the parent issue for now: https://github.com/Kampfkarren/selene/issues/469#issuecomment-2016543458

This pr will still be very helpful if the issue gets reopened in the future. If that happens, this pr will also need to update the vscode extension with the new config resolution.

Zeioth:
Seriously

docs/src/cli/usage.md
```diff
@@ -16,7 +16,7 @@ FLAGS:
 
 OPTIONS:
         --color <color>                     [default: auto]  [possible values: Always, Auto, Never]
-        --config <config>                  A toml file to configure the behavior of selene [default: selene.toml]
+        --config <config>                  A toml file to configure the behavior of selene [default: .selene.toml]
```
chriscerie, line 19: This isn't technically accurate because the default is both `selene.toml` and `.selene.toml`.

selene/src/main.rs
```diff
@@ -391,6 +391,55 @@ fn read_file(checker: &Checker<toml::value::Value>, filename: &Path) {
     );
 }
 
+/// Reads a config file and return content and path. None if not found or empty.
+fn read_config_file() -> Option<(String, Option<PathBuf>)> {
```
chriscerie, line 395: Why the nested option? There shouldn't be a case where a config's contents exist but not its path.

selene/src/main.rs
```diff
@@ -391,6 +391,55 @@ fn read_file(checker: &Checker<toml::value::Value>, filename: &Path) {
...
+        if let Ok(contents) = fs::read_to_string(path) {
+            config_contents = contents;
+            config_path = Some(PathBuf::from(path));
+            break;
+        }
+    }
+
+    if config_contents.is_empty() {
+        return None;
+    }
```
chriscerie, line 410: Empty config should be handled separately from missing config.

selene/src/main.rs
```diff
@@ -391,6 +391,55 @@ fn read_file(checker: &Checker<toml::value::Value>, filename: &Path) {
     );
 }
 
+/// Reads a config file and return content and path. None if not found or empty.
+fn read_config_file() -> Option<(String, Option<PathBuf>)> {
+    let config_paths_to_check = [".selene.toml", "selene.toml"];
```
chriscerie, line 396: Should be a const.

selene/src/main.rs
```diff
@@ -391,6 +391,55 @@ fn read_file(checker: &Checker<toml::value::Value>, filename: &Path) {
...
+    let mut config_contents = String::new();
+    let mut config_path = None;
+
+    for path in &config_paths_to_check {
+        if let Ok(contents) = fs::read_to_string(path) {
+            config_contents = contents;
+            config_path = Some(PathBuf::from(path));
+            break;
+        }
```
chriscerie, line 405: You're discarding `fs::read_to_string` error messages here. Returning the function as an err result would make more sense than option.

selene/src/main.rs
```diff
@@ -412,17 +461,8 @@ fn start(mut options: opts::Options) {
...
-                let config_contents = match fs::read_to_string(config_path) {
-                    Ok(contents) => contents,
-                    Err(error) => {
-                        error!("Error reading config file: {error}");
-                        std::process::exit(1);
-                    }
-                };
-
-                (config_contents, config_path)
+                let (config_content, config_path) = parse_config_file_as_string().unwrap();
```
chriscerie, line 464: Unwrap can panic. Need an error message here.

selene/src/main.rs
```diff
@@ -391,6 +391,55 @@ fn read_file(checker: &Checker<toml::value::Value>, filename: &Path) {
...
+    }
+
+    if config_contents.is_empty() {
+        return None;
+    }
+
+    Some((config_contents, config_path))
+}
+
+/// Read the config file and return it as `Option<(String, &'static Path)>`.
```
chriscerie, line 415:
```suggestion
```

selene/src/main.rs
```diff
@@ -391,6 +391,55 @@ fn read_file(checker: &Checker<toml::value::Value>, filename: &Path) {
...
+            return Some((config_contents, Box::leak(Box::new(path)).as_path()));
+        }
+    } else {
+        error!("Error reading config file");
+        std::process::exit(1);
+    }
+    None
+}
+
+/// Read the config file and return it as `CheckerConfig<toml::Value>, Option<PathBuf>`.
```
chriscerie, line 428:
```suggestion
```

chriscerie (commented):
Also need relevant changes to vscode extension.

selene/src/main.rs
Zeioth: Closed in https://github.com/Kampfkarren/selene/pull/591/commits/822bc446494e5e856a9ae5df5ed0fd56fd9e0ebe

selene/src/main.rs
Zeioth: I cannot repro the scenario in which this provoke panic.  All scenarios I've tried are successful:

* `.selene.toml` exists, and it is used.
*  `selene.toml` exists, and it is used.
* `config file` doesn't exist (defaults are loaded correctly).
* config file exists but some option is formatted incorrectly, so a notification is sent to the user, and exit (for `.selene.toml` and `selene.toml`).
* config exists but it's empty, so defaults are loaded (for `.selene.toml` and `selene.toml`).

Zeioth:
[chriscerie](https://github.com/chriscerie) since there are no bugs, the code has already being refactored as specified, and all the new reviews are stylistic preferences, or would affect parts outside the PR, I'm gonna leave the code as it is. Feel free to co-author if you see something you would prefer to do in a different way.

The code is safe to merge.

selene/src/main.rs
Zeioth: We silence that error because if no file has been found, we return None. That way we can treat the lack of config file it as we decide for each case inside `parse_config_file_as_string()`, or `parse_config_file_as_checkerconfig()`.

```rust
    for path in &CONFIG_PATHS_TO_CHECK {
        if let Ok(contents) = fs::read_to_string(path) {
            config_contents = contents;
            config_path = Some(PathBuf::from(path));
            break;
        }
    }

    if config_contents.is_empty() {
        return None;
    }

    Some((config_contents, config_path))
```

selene/src/main.rs
Zeioth: Please go ahead if you have a better solution.

chriscerie: This doesn’t answer my question. What would it mean if it returns `Some(String, None)`? Did you intend to return `Option<String, PathBuf>` instead?

selene/src/main.rs
chriscerie: This prevents other errors from propagating, like trying to read invalid UTF-8 characters. I’d check if the path exists first so you can handle that separately.

selene/src/main.rs
Zeioth: The PR depends on the existing code. This is needed to satisfy the current architecture. Feel free to improve it if you think you can.

chriscerie:
I appreciate your contribution, but this pr can't be merged until these bugs are fixed (see above comments for root causes). I'm happy to step in and co-author to get this merged but as there are other selene features with higher priority, I make no promises on the timeline.

### Config with invalid utf-8 character
Currently
image
New
image

### Missing config
Currently
image
New
image

selene/src/main.rs
Zeioth: Solved in https://github.com/Kampfkarren/selene/pull/591/commits/ade463c78eca904e2d3fd801c3f6a792f258edc6

selene/src/main.rs
Zeioth: My bad, this was actually fixable. Solved in https://github.com/Kampfkarren/selene/pull/591/commits/ade463c78eca904e2d3fd801c3f6a792f258edc6

selene/src/main.rs
Zeioth: fixed in https://github.com/Kampfkarren/selene/pull/591/commits/ade463c78eca904e2d3fd801c3f6a792f258edc6

selene/src/main.rs
Zeioth: fixed in https://github.com/Kampfkarren/selene/pull/591/commits/ade463c78eca904e2d3fd801c3f6a792f258edc6

Zeioth:
bad UTF-8 error is controlled now.
screenshot_2024-03-27_14-18-16_067338925

Zeioth:
Missing file is also parsed correctly (this is my build, not the package manager one).
screenshot_2024-03-27_14-34-02_956661142

Zeioth:
Relevant changes to vscode extension https://github.com/Kampfkarren/selene/pull/591/commits/47b06421a96fca5d0f3646112ded08b01d4b8714

@chriscerie It should be ready but I have no clue how to test this on vscode. Could you test it or tell me how you test it so I can try to reproduce?

chriscerie (changes requested):
Missing config

Before
image
Now
image

Permission denied

Before
image
Now
image

selene/src/main.rs
```diff
@@ -391,6 +390,58 @@ fn read_file(checker: &Checker<toml::value::Value>, filename: &Path) {
...
+        match CONFIG_PATHS_TO_CHECK
+            .iter()
+            .find_map(|path| {
+                match fs::read(path) {
+                    Ok(contents) => {
+                        if let Ok(valid_str) = String::from_utf8(contents) {
+                            Some((valid_str, PathBuf::from(path)))
+                        } else {
+                            // Handle invalid UTF-8
+                            error!("Error reading config file: {}, stream did not contain valid UTF-8", path);
```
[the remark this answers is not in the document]
Zeioth: Closed in https://github.com/Kampfkarren/selene/pull/591/commits/4ca19d44361287c44dcbe6626136f1993f4703f9

screenshot_2024-05-07_18-13-41_834796775
Same, but printed by neovim.
screenshot_2024-05-07_18-13-16_772390668

selene/src/main.rs
```diff
@@ -391,6 +390,58 @@ fn read_file(checker: &Checker<toml::value::Value>, filename: &Path) {
...
+                            std::process::exit(1);
+                        }
+                    }
+                    Err(_) => None, // if a file do not exist, return none
```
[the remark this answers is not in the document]
Zeioth: @chriscerie thank you for being so explicit in your answer. It should be fixed in 
https://github.com/Kampfkarren/selene/pull/591/commits/aff6f9a7a0719df393badf4a82442b5f519771ce
https://github.com/Kampfkarren/selene/pull/591/commits/7c844a3d5a91b680a9d20c17665e7b6544ba2c64

Every manual test I've performed seem to work as expected. Including permission denied.
screenshot_2024-05-08_21-21-31_356340982

Bad UTF-8
screenshot_2024-05-08_22-13-15_154181796

## About the missing config error
We probably don't want to display the `missing config` error, because:
* The error would also display when using selene normally (not only when doing `selene validate-config`).
* It's not really an error, as selene will fallback to the default config.
* The error would display in the code editor, which is something we prob don't want.

A workaround we could use is handling the error with `println!` instead of `error!` so it prints the error when using selene on the terminal, but not on code editors. I've tested it and it works, but for now I'm just gonna silence the error unless you tell me otherwise. If you decide to enable it would look like:

screenshot_2024-05-08_22-22-58_734164313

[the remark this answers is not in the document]
chriscerie: > The error would also display when using selene normally (not only when doing selene validate-config).

The existing behavior for missing config with `validate-config` is to error which behaves differently from the main command, and that should not be changed. The two commands are handled separately. See https://github.com/Kampfkarren/selene/blob/5c02fe9b9e25137b5896dac5e941854353cdb200/selene/src/main.rs#L456-L457

> It's not really an error, as selene will fallback to the default config.

It's not intuitive to me for the command to fallback to the default config when its purpose is to validate the user's config.

> The error would display in the code editor, which is something we prob don't want.

Where would it display in the code editor?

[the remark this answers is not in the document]
Zeioth: @chriscerie Ok the last commit should solve it:

```rust
let error = io::Error::new(
  io::ErrorKind::NotFound,
 "Error reading config file: No such file or directory (os error 2)"
);
error!("{}", error);
```

It displays on `./selene validate-config`
screenshot_2024-05-21_15-35-42_777689703

All the border cases seem to work fine too:
* The error display when running when running validate-config if none of the files ./.selene or ./selene is present.
* The error doesn't display on the code editor, only on validate-config.
* The error doesn't display when running validate-config if ./.selene or ./selene exist are the file is correct.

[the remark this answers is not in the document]
Zeioth: Ok I solved all border cases. Ready to review.

PR #599, opened 2024-05-17
add aarch64 build

joske:
Hi,

In asahi linux (aarch64), selene is not available in neovim because mason can not download an artifact for it (there is no aarch64 linux on github releases). To fix this, I added a cross compilation step to generate and upload an aarch64 linux binary.

I tested this in my fork, and it correctly generates aarch64 binaries.

Please let me know if you like something changed!

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -78,6 +78,18 @@ jobs:
       with:
         name: selene-linux
         path: ./target/release/selene
+    - name: Build arm64 (Default features)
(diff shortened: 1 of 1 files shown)


joske:
@Kampfkarren @chriscerie bump?

Kampfkarren:
Apologies! Haven't been able to put too much time into selene.

joske:
@Kampfkarren Great! If you could now make a new release so mason can actually download the new arch  :grin:

PR #605, opened 2024-06-21
add Path2DControlPoint.new to roblox_base.yml

Stenimated:
this lint works fine for the majority. besides this bellow, it'll require all 3 args if the second one is present 
Path2DControlPoint.new(UDim2.new(), UDim2.new()) -- RUNTIME ERROR
any way to lint for that

### DIFF
--- a/selene-lib/default_std/roblox_base.yml
+++ b/selene-lib/default_std/roblox_base.yml
@@ -349,15 +349,6 @@ globals:
   OverlapParams.new:
     args: []
     must_use: true
-  PathWaypoint.new:
-    args:
-      - required: false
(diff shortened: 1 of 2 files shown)


selene-lib/default_std/roblox_base.yml
```diff
@@ -358,6 +358,18 @@ globals:
         type:
           display: PathWaypointAction
     must_use: true
+  Path2DControlPoint.new:
```
chriscerie, line 361: Should precede PathWaypoint.new.

chriscerie (approved):
Thanks!

PR #607, opened 2024-06-26
Amend installation.md so as to make its instruction about `--no-default-features` harder to misunderstand

### LINKED ISSUE #606: Problematic documentation for `--no-default`
`selene --help` (within selene 0.27.1, on Linux) does not mention the `--no-default` flag that is mentioned [here](https://kampfkarren.github.io/selene/cli/installation.html). At that latter location, though, the flag is called

   --no-default- [_sic_]

Is that latter, with the extra dash, correct? In either version of the syntax, I cannot get the command working:

```
$ selene --no-default  ~/scripts/conky/lua
error: Found argument '--no-default' which wasn't expected, or isn't valid in this context

USAGE:
    selene [FLAGS] [OPTIONS] <files>...
    selene <SUBCOMMAND>

For more information try --help
$ selene --no-default- ~/scripts/conky/lua
error: Found argument '--no-default-' which wasn't expected, or isn't valid in this context

USAGE:
    selene [FLAGS] [OPTIONS] <files>...
    selene <SUBCOMMAND>

For more information try --help
```

LinuxOnTheDesktop:
Closes #606.

### DIFF
--- a/docs/src/cli/installation.md
+++ b/docs/src/cli/installation.md
@@ -16,4 +16,4 @@ cargo install --branch main --git https://github.com/Kampfkarren/selene selene
 ```
 
 ### Disabling Roblox features
-selene is built with Roblox specific lints by default. If you don't want these, type `--no-default-features` after whichever command you choose.
+selene is built with Roblox specific lints by default. If you don't want these, then pass `--no-default-features` to the `cargo install` command.


Kampfkarren:
Thanks!

PR #616, opened 2024-09-22
Fix failing build due to deprecated action

chriscerie:


Kampfkarren:
Pulling this into #610

PR #618, opened 2024-10-12
Fix: Exclude not working with absolute path in some cases (#593)

### LINKED ISSUE #593: Exclude not working with absolute path in some cases
These are all executed from the `selene-test` workspace.

Works:
```
exclude = ["**/file.lua"]
```
```
selene C:\Users\chris\OneDrive\Documents\selene-test\src\file.lua
```

Works:
```
exclude = ["src/**"]
```
```
selene src/file.lua
```

Doesn't work:
```
exclude = ["src/**"]
```
```
selene C:\Users\chris\OneDrive\Documents\selene-test\src\file.lua
```

dqkqd:
Fix #593 

Exclude files if they match exclude patterns using **working directory** (config directory if a config file is provided, otherwise it is the current directory) as a base.

I'm not sure should I add testcases for this, since this requires refactor into smaller functions.


selene/src/main.rs
```diff
@@ -636,6 +643,26 @@ fn start(mut options: opts::Options) {
...
+    // If user provides a config file, we use that as our working directory, otherwise fallback to the current directory.
+    // The working directory needs to be an absolute path to match with paths in different directories.
+    let working_directory = config_directory.or_else(|| current_dir.canonicalize().ok());
+
+    let is_excluded = |path: &Path| -> bool {
+        if options.no_exclude {
+            return false;
+        }
+
+        // File pattern between the path we are checking and current working directory.
```
chriscerie, line 656:
```suggestion
```

chriscerie (commented):
Thanks! Need changelogs for:
- the fix
- changing undocumented behavior that exclude paths are now relative to the config file instead of where selene is invoked

We should also properly document that in https://kampfkarren.github.io/selene/usage/configuration.html#excluding-files-from-being-linted

chriscerie:
We should keep existing behavior if changing is not required to fix the issue. Changing it should have its own issue and discussion.

dqkqd:
Thank @chriscerie 

> We should keep existing behavior if changing is not required to fix the issue. Changing it should have its own issue and discussion.

Then, I use the current directory (where selene is invoked) to convert the absolute path to relative path.

I updated the PR and the [PR description](https://github.com/Kampfkarren/selene/pull/618#issue-2582604835) as well. Please have a look

Issue #645, opened 2025-07-22
Lua 5.2+ features are disabled


khvzak:
The standard `selene` binary is built without all `selene-lib` features except `roblox`.

There is no way to enable Lua 5.3+ stdlib through `selene.toml`, despite the [documentation says] it's supported.

[documentation says]: https://kampfkarren.github.io/selene/usage/std.html#base

marek22k:
Is there a current workaround? I get `lua version lua52 in standard library, but feature for it is not enabled`

nurpax:
> Is there a current workaround? I get `lua version lua52 in standard library, but feature for it is not enabled`

I never found a workaround.  IMO my PR fixes it correctly.

Issue #340, opened 2022-02-08
Support for Lua 5.3


sanga:
Feel free to close if you wish as I presume this is a nontrivial request. We're a lua 5.3 house, so I'd appreciate support for it.

Kampfkarren:
Needs full moon support, but possible outside of that.

sanga:
Noting here that full-moon now has 5.3 support (https://github.com/Kampfkarren/full-moon/issues/221)

shenef:
As i understand it, this is not blocked by full-moon anymore? 
I'd appreciate an update on this, currently using bitwise operators results in a parse error.
![image](https://github.com/Kampfkarren/selene/assets/20268890/71fcace8-feaa-42ea-a62a-f46f5e3c9901)

PR #646, opened 2025-07-27
fix: extend() method now merges lua_versions instead of replacing

nurpax:
Previously, lua53.yml would lose Lua 5.3 support when extending lua52.yml, causing bitwise operators to fail parsing.

for #340, #645

### DIFF
--- a/selene-lib/src/standard_library/mod.rs
+++ b/selene-lib/src/standard_library/mod.rs
@@ -320,9 +320,9 @@ impl StandardLibrary {
                 }),
         );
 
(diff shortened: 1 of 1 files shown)


nurpax:
Hi!  I rebased the PR to the latest Selene main.

I think this change is a valuable change to anyone using selene CLI on Lua 5.3 or 5.4 codebase with their own std library extensions.  The currently released version forces earlier than Lua 5.3 version despite what's configured in the selene.toml file.

The problem is well explained here: https://github.com/Kampfkarren/selene/issues/645

Could you please consider merging this fix?

PR #655, opened 2025-10-29
Fix buffer writebits and readbits

TheAlizabeth:
Fixes writebits and readbits methods for buffers.


chriscerie:
Need changelog

TheAlizabeth:
My bad.

PR #666, opened 2026-04-01
fix: forward Lua version features from selene-lib to CLI crate

### LINKED ISSUE #581: Using the LuaJIT extension syntax for ULL numbers causes a parse error (despite already being supported?)
This is a follow-up (of sorts) to https://github.com/Kampfkarren/full-moon/issues/255

---

As I was looking into adopting selene for use with my project, I discovered that it fails to parse the following LuaJIT-supported syntax:

```lua

local wgpu = {
	WGPU_ARRAY_LAYER_COUNT_UNDEFINED = 0xffffffffULL,
	WGPU_COPY_STRIDE_UNDEFINED = 0xffffffffULL,
	WGPU_LIMIT_U32_UNDEFINED = 0xffffffffULL,
	WGPU_LIMIT_U64_UNDEFINED = 0xffffffffffffffffULL,
	-- ... rest omitted for brevity
}
```

Running `selene` on this with `base: lua51` in the config fails with a parser error:

```sh
error[parse_error]: unexpected token `{`
  ┌─ Runtime/Bindings/webgpu.lua:3:14
  │
3 │ local wgpu = {
  │              ^ expected expression
```

I've tried changing the base to `lua51+lua52` and just `lua52` as well - to no effect. This was using the latest git checkout of selene.

---

It seems that `selene` uses a recent version of the parser (according to the cargo file), which should include support for the extension. Is there anything in particular that I need to do in order to enable this feature? The PR mentioned `lua52` so that's what I tried. FWIW, StyLua seems to take no issue with the code and it's using the same parser, so I'm a bit confused as to why selene choked on the file :)

roobie:
The selene-lib crate defines lua52, lua53, lua54, and luajit as default features (enabling the corresponding full_moon parser features), but the selene CLI crate imports selene-lib with default-features = false and only re-exports the roblox feature. This means cargo install selene and the GitHub release binaries cannot parse Lua 5.2+ syntax (goto/labels), LuaJIT extensions (ULL number suffixes), or \x hex string escapes.

Forward all four Lua version features through the CLI crate and include them in its default feature set, matching selene-lib's own defaults.

Fixes #581

This PR was assisted by AI

### DIFF
--- a/selene/Cargo.toml
+++ b/selene/Cargo.toml
@@ -40,6 +40,10 @@ ureq = { version = "2.6.2", features = ["json"], optional = true }
 pretty_assertions = "1.3"
 
 [features]
-default = ["roblox"]
+default = ["lua52", "lua53", "lua54", "luajit", "roblox"]
 tracy-profiling = ["profiling/profile-with-tracy", "tracy-client"]
+lua52 = ["selene-lib/lua52", "full_moon/lua52"]
+lua53 = ["selene-lib/lua53", "full_moon/lua53"]
+lua54 = ["selene-lib/lua54", "full_moon/lua54"]
+luajit = ["selene-lib/luajit", "full_moon/luajit"]
 roblox = ["selene-lib/roblox", "full_moon/roblox", "ureq"]


nurpax:
I noticed a weird subtlety related to this.

I've sent this PR https://github.com/Kampfkarren/selene/pull/646 which fixed the problem for me (the propagation of lua version).  However, I now stumbled upon your change and I started wondering why did my change fix the problem for me, when your change looks like it should be required too.

Well, I was building my selene binary with `cargo build --release` and not with `cargo install selene` (or running `cargo run -p selene`).  If building selene with `cargo build` it seems like the defaults are picked for the full workspace which then sets more defaults than the `selene/Cargo.toml`.  

I verified this theory with the below repros:

```
 # selene.toml
std = "lua53"

-- test.lua
local x = 1 & 2
```

running `cargo run -p selene -- --config selene.toml test.lua` fails with 

```
ERROR: lua version lua53 in standard library, but feature for it is not enabled
ERROR: lua version lua52 in standard library, but feature for it is not enabled
error[parse_error]: unexpected character &
  ┌─ test.lua:1:13
  │
1 │ local x = 1 & 2
  │             ^

error[parse_error]: unexpected token `2`
  ┌─ test.lua:1:15
  │
1 │ local x = 1 & 2
  │               ^ unexpected token, this needs to be a statement
```

but it works fine if I instead do:

```
cargo build --release
../target/release/selene --config selene.toml test.lua
```

So IMO both PRs are needed.  Mine fixes Lua version inheritance in config files and yours adds the right build time feature enables.

PR #676, opened 2026-06-18
Preserve standard-library metadata across simple local aliases of tables

### LINKED ISSUE #675: Standard-library metadata is lost when a standard-library table is assigned to a local
**Repro**

`zig.yml`

```yml
globals:
  z2.ecs.create_entity:
    args: []
    must_use: true
```

`selene.toml`
```toml
std = "zig"
```

`repro.lua`

```lua
local ecs = z2.ecs
local e = ecs.create_entity(3) -- `z2.ecs.create_entity` requires 0 parameters and 1 was passed
```

**Expected behavior**

Selene should report something like: `Selene should report that `z2.ecs.create_entity` requires 0 parameters and 1 was passed` (because `ecs` is just a local alias of `z2.ecs`.)

This applies to standard-library definitions, including libraries that I define in my own yml files.

**What actually happens**

No standard-library warning is reported once the table is assigned to a local:

```lua
local ecs = z2.ecs
local e = ecs.create_entity(3)
```

The direct form does warn:

```lua
local e = z2.ecs.create_entity(3)
```

**Suggested fix**

I'm working on a PR for this.

nurpax:
Calls via local aliases of tables did not lint correctly.  For example, given the below code with warnings:

```
  math.max(1,3)    -- unused return value of `math.max` must be used
```
But if you call max via a local alias:

```
  local math2 = math
  math2.max(1,3)   -- no warning at all
```

In Lua codebases with nested namespaces, this becomes a problem, because Selene's linting ability is completely lost on code like below:

```
  local ecs = z2.ecs   -- grab ecs submodule from z2
  local e = ecs.create_entity(3) -- selene doesn't know about create_entity
```

This commit beefs up selene to track standard-library metadata tracking via such aliases as follows:

1. When Selene sees an assignment from a known path like `z2.ecs` or `math`, it stores that origin on the assigned variable.

   Example: after `local ecs = z2.ecs`, Selene records that `ecs` is an alias of the path `z2.ecs`.

2. When Selene later sees a call or field access through that variable, it tries to recover the original path before doing standard-library checks.

   Example: `ecs.create_entity(3)` can be treated like `z2.ecs.create_entity(3)`.

3. Selene only does that for variables that are known aliases of an existing path, not for arbitrary locals.

   Example: `local z2 = {}; local ecs = z2` should stay local data, not accidentally turn into the standard-library `z2.ecs`.

LLM disclosure: I did use codex in development of this change.  I did however iterate on this a fair bit, reviewing, writing tests, improving test coverage, and IMO the feature is well-tested and high quality.  I'm not a professional Rust coder though, so I wouldn't be surprised if the Rust code could be improved even further.

fixes #675

### DIFF
--- a/selene-lib/src/ast_util/name_paths.rs
+++ b/selene-lib/src/ast_util/name_paths.rs
@@ -44,7 +44,7 @@ pub fn name_path(expression: &ast::Expression) -> Option<Vec<String>> {
                 name_path_from_prefix_suffix(expression.prefix(), expression.suffixes())
             }
 
-            ast::Var::Name(name) => Some(vec![name.to_string()]),
+            ast::Var::Name(name) => Some(vec![name.token().to_string()]),
 
             _ => None,
         }
--- a/selene-lib/src/ast_util/scopes.rs
+++ b/selene-lib/src/ast_util/scopes.rs
@@ -13,7 +13,7 @@ use id_arena::{Arena, Id};
 
 use crate::ast_util::extract_static_token;
 
-use super::expression_to_ident;
+use super::{expression_to_ident, name_paths::name_path};
 
 type Range = (usize, usize);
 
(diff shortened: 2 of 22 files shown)


nurpax:
BTW, the code does not attempt to track writes to standard library entries through aliases.  E.g., something like below won't work quite right:

```
local math2 = math
math2.max = custom_max
math2.max(x, y)     -- this line will lint as if selene saw math.max
```

I wanted to draw the line at something simple that still brings a lot of value in common Lua code.  A useful future update might be to warn if a write to standard library tables is done through an alias, but otherwise stop at that.

